PR #2, opened 2022-03-08
Update statement

IvanRamyk:


app/players.json
```diff
@@ -0,0 +1,62 @@
...
+    }
+  },
+  "nick": {
+    "email": "nick@gmail.com",
+    "bio": "Hello, I'm Nick",
+    "race": {
+      "name": "human",
+      "description": "Human race"
+    },
+    "guild": {
```
danylott, line 57: Maybe make guild empty for someone - as soon as it can be null

app/players.json
```diff
@@ -0,0 +1,62 @@
...
+      "name": "sailors",
+      "description": "A community of the sailors"
```
danylott, line 59: Same with description somewhere

app/players.json
```diff
@@ -0,0 +1,62 @@
+{
+  "john": {
+    "email": "john@gmail.com",
+    "bio": "Hello, I'm John, elf ranger",
+    "race": {
+      "name": "elf",
+      "description": "The magic race"
```
danylott, line 7: What about Skill for race?
Just nothing with it?

danylott (changes requested):
Add tests and reference solution please

danylott (approved):

PR #3, opened 2022-03-12
Solution

Krimat:


main.py
```diff
@@ -1,10 +1,59 @@
 import init_django_orm  # noqa: F401
+import json
+
 
 from db.models import Race, Skill, Player, Guild
 
 
+def create_guild(data: dict):
+    if 'description' in data :
```
MasterpieceElbow, line 9: You don't need this condition. You can use the dict method `get` with the default value `None`, because guild description can be `null`.

main.py
```diff
@@ -1,10 +1,59 @@
...
+        print(data['name'])
```
MasterpieceElbow, line 10: Delete this output.

main.py
```diff
@@ -1,10 +1,59 @@
...
+    for raw_player in raw_players:
+        player_data = raw_players[raw_player]
+        race_data = player_data['race']
+        skills_data = race_data['skills']
+        guild_data = player_data['guild']
+        race = Race.objects.get_or_create(
+            name=race_data['name'],
+            description=race_data['description']
+        )[0]
+        race.save()
```
MasterpieceElbow, line 34: You don't need to `save()` after the `get_or_create`.

MasterpieceElbow (changes requested):

main.py
```diff
@@ -1,10 +1,48 @@
...
+        if guild_data is not None:
+            guild = create_guild(guild_data)
+        player = Player.objects.create(
+            nickname=raw_player,
+            email=player_data['email'],
+            bio=player_data['bio'],
+            race=race,
+            guild=guild,
+        )
+        player.save()
```
MasterpieceElbow, line 45: You don't need `.save()` after `objects.create()`

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #4, opened 2022-03-12
'Solution'

liuda-chuzdiuk:


main.py
```diff
@@ -1,10 +1,72 @@
...
+    Player.objects.all().delete()
+    Guild.objects.all().delete()
+
+    with open("players.json") as file:
+        players_data = json.load(file)
+
+    races = [players_data[element]["race"] for element in list(players_data)]
+    guilds = [players_data[element]["guild"] for element in list(players_data)]
+
+    for race in races:
```
Krimat, line 20: Also, you can use get_or_create() method to create object if does not exist(just for info)

main.py
```diff
@@ -1,10 +1,72 @@
...
+
+    for guild in guilds:
+        if guild is not None and not Guild.objects.filter(
+                name=guild["name"]).exists():
+            Guild.objects.create(
+                name=guild["name"],
+                description=guild["description"]
+            )
+
+    Skill.objects.create(
```
Krimat, line 35: Why literals used here?

Krimat (changes requested):

vldtsw (approved):

MasterpieceElbow (approved):

PR #11, opened 2022-03-14
'Solution'

serhii56465:


main.py
```diff
@@ -1,10 +1,51 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main():
-    pass
+    with open("players.json") as game_data:
+        players_data = json.load(game_data)
+    for players in players_data:
+        if Race.objects.filter(name=players_data[players]
+           ["race"]["name"]).exists() is False:
```
Nazar-kob, line 12: you can use here "not" instead of "is False", but I approve

main.py
```diff
@@ -1,10 +1,51 @@
...
+                                      name=players_data[players]
+                                      ["guild"]["name"]))
+        elif players_data[players]["guild"] is None:
+            Player.objects.create(nickname=players,
+                                  email=players_data[players]["email"],
+                                  bio=players_data[players]["bio"],
+                                  race=Race.objects.get(
+                                      name=players_data[players]
+                                      ["race"]["name"]),
+                                  guild=None)
```
Nazar-kob, line 48: you repeat "Player.create" twice. I think this is the wrong style of writing code. You can create a value such as "get_player_guild" and check the value. Then use the value in "Player.create"!

Nazar-kob (changes requested):

PR #13, opened 2022-03-14
Develop 2

vldtsw:


db/models.py
```diff
@@ -1 +1,35 @@
...
+class Player(models.Model):
+    nickname = models.CharField(max_length=255)
+    email = models.CharField(max_length=255)
+    bio = models.CharField("Stores a short description provided by a "
+                           "user about himself/herself",
+                           max_length=255)
+    race = models.ForeignKey(Race,
+                             on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild,
+                              on_delete=models.CASCADE,
```
phantom-terminal, line 33: I think it's better to use on_delete=models.SET_NULL, because players shouldn't be deleted when guild is deleted. And the player's guild would be set to null

db/models.py
```diff
@@ -1 +1,35 @@
...
+                              null=True)  # TODO null?
```
phantom-terminal, line 34: It's better to remove the commentary.

main.py
```diff
@@ -1,10 +1,52 @@
...
+        players_data = players[player]
+
+        #  guild
+        if players_data["guild"] is not None:
+            if not Guild.objects.filter(
+                    name=players_data["guild"]["name"]).exists():
+                Guild.objects.create(
+                    name=players_data["guild"]["name"],
+                    description=players_data["guild"]["description"])
+                players_guild = Guild.objects.get(
```
phantom-terminal, line 22: This line should be out of the "if" statement so you can get the guild object when it already exists.

phantom-terminal (changes requested):

main.py
vldtsw: Thanks! Fixed

phantom-terminal (approved):
Good job)

MasterpieceElbow (approved):

PR #15, opened 2022-03-15
Solution

phantom-terminal:


main.py
```diff
@@ -1,10 +1,60 @@
...
+        race_data = players[player]["race"]
+        guild_data = players[player]["guild"]
+
+        if not Race.objects.filter(name=race_data["name"]).exists():
+            Race.objects.create(
+                name=race_data["name"],
+                description=race_data["description"]
+            )
+
+        race_name = Race.objects.get(name=race_data["name"])
```
TPolina, line 32: Better use just "race" as a variable name (or something else). "race_name" is a bit confusing as you get not a race_name but the race object

main.py
```diff
@@ -1,10 +1,60 @@
...
+                    race=race_name
+                )
+
+        if guild_data:
+            if not Guild.objects.filter(name=guild_data["name"]).exists():
+                Guild.objects.create(
+                    name=guild_data["name"],
+                    description=guild_data["description"]
+                )
+                guild_name = Guild.objects.get(name=guild_data["name"])
```
TPolina, line 48: this line should be out of if statement so you can get guild object when it already exists.

main.py
```diff
@@ -1,10 +1,60 @@
...
+        else:
+            guild_name = None
```
TPolina, line 50: Better use just "guild" or "guild_object" as a variable name

main.py
```diff
@@ -1,10 +1,60 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main():
-    pass
+    global guild_name
```
TPolina, line 9: why do you need this?

TPolina (changes requested):

db/models.py
```diff
@@ -1 +1,33 @@
...
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255)
+    bonus = models.CharField("This field describes "
+                             "what kind of bonus "
+                             "players can get from it.",
+                             max_length=255)
```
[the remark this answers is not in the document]
phantom-terminal: Thx)

db/models.py
```diff
@@ -1 +1,33 @@
...
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255)
+    email = models.CharField(max_length=255)
+    bio = models.CharField("This field to store "
+                           "a short description "
+                           "provided by a user "
+                           "about himself/herself.",
```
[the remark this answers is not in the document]
phantom-terminal: I fixed it

db/models.py
```diff
@@ -1 +1,33 @@
...
+                           max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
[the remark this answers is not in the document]
phantom-terminal: Thx, I agree with you

main.py
phantom-terminal: Thanks, I had to make a global "guild_name" because of that. I fixed that.

TPolina (approved):
Nice)

MasterpieceElbow (approved):

PR #16, opened 2022-03-15
Solution

TPolina:


db/models.py
```diff
@@ -1 +1,26 @@
...
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255)
+    description = models.TextField(null=True)
```
phantom-terminal, line 17: "Guild description - a text field, can be blank". In addition to what may be null can be blank

phantom-terminal (changes requested):

phantom-terminal (approved):

MasterpieceElbow (approved):

PR #20, opened 2022-04-21
Solution

anitaposliedova:


alexandra-halytska (approved):
try to do the same but using functions like: fill_race, fill_skill, fill_guild, in this way will be less code

main.py
```diff
@@ -1,10 +1,62 @@
 import init_django_orm  # noqa: F401
-
+import json
 from db.models import Race, Skill, Player, Guild
 
 
 def main():
```
IvanRamyk, line 6: The logic of this function is better divided into several functions.
For example, a separate function for creating a guild/race.

IvanRamyk (changes requested):

MasterpieceElbow (approved):

PR #21, opened 2022-06-05
v1

SairusQw:


IvanRamyk (changes requested):
Tests are failed because no migrations provided.
Django should know the schema of DB while testing

PR #24, opened 2022-06-16
Solution

YuriiKindrat:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField()
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
Magus6784, line 25: it's better to use SET_NULL because if we delete a guild, ALL players in that guild will be deleted.

main.py
```diff
@@ -1,10 +1,50 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main():
-    pass
+
+    with open("players.json") as f:
+        data = json.load(f)
+
+    for name in data:
```
Magus6784, line 12: better use sctructure
name, information in data.values()
No need to use data[name], just information(or another name)

main.py
```diff
@@ -1,10 +1,50 @@
...
+
+        # check if race exists
+        if Race.objects.filter(name=data[name]["race"]["name"]).exists():
+            race = Race.objects.get(name=data[name]["race"]["name"]).id
+        else:
+            Race.objects.create(name=data[name]["race"]["name"],
+                                description=data[name]["race"]["description"])
```
Magus6784, line 19: You can use very comfortable method get_or_create
So you can merge get and create

Magus6784 (changes requested):

Magus6784 (approved):

MasterpieceElbow (approved):

MasterpieceElbow (approved):

PR #25, opened 2022-06-16
Solution

Serhii-Leonenko:


Serhii-Leonenko:
> There are new requirement for names. They should be unique. And try to split into several functions. S from SOLID ;)

oh, new requirment, cool ))

main.py
```diff
@@ -1,10 +1,101 @@
...
-    pass
+    players_data = parse_file()
+
+    for nickname, data in players_data.items():
+
+        race_name = data["race"]["name"]
+        race_description = data["race"]["description"]
+        create_race(race_name, race_description)
+
+        race = Race.objects.get(name=race_name)
```
rotsen18, line 82: For me its better to add return race in create_race func and then you won't need to make another query to db

rotsen18 (approved):
Now its look much more better, good job

main.py
Serhii-Leonenko: agree, thanks)

MasterpieceElbow (approved):

PR #33, opened 2022-06-20
Solution

BohdanKryven:


main.py
```diff
@@ -1,10 +1,86 @@
...
+                                          name=value["race"]["name"]),
+                                      guild=None)
+
 
 def main():
-    pass
+    Race.objects.all().delete()
+    Skill.objects.all().delete()
+    Guild.objects.all().delete()
+    Player.objects.all().delete()
```
rotsen18, line 79: You don't need to clear all models

rotsen18 (changes requested):

main.py
BohdanKryven: fixed. thanks. this was for test runs

main.py
```diff
@@ -1,10 +1,81 @@
...
+                                          name=value["race"]["name"]),
+                                      guild=Guild.objects.get(
+                                          name=value["guild"]["name"]))
+            else:
+                Player.objects.create(nickname=key,
+                                      email=value["email"],
+                                      bio=value["bio"],
+                                      race=Race.objects.get(
+                                          name=value["race"]["name"]),
+                                      guild=None)
```
rotsen18, line 72: How about a little simplification?
```suggestion
guild = None
if value["guild"] is not None:
    guild = Guild.objects.get(name=value["guild"]["name"])
Player.objects.create(nickname=key,
                      email=value["email"],
                      bio=value["bio"],
                      race=Race.objects.get(
                          name=value["race"]["name"]),
                      guild=guild)
```

rotsen18 (approved):

main.py
BohdanKryven: Smart solution! It's probably better than mine

main.py
```diff
@@ -1,10 +1,75 @@
...
+PLAYERS = "players.json"
+
+
+def load_data_json():
+    with open(PLAYERS) as file:
+        return json.load(file)
+
+
+def get_race_info():
+    return [value["race"] for key, value in load_data_json().items()]
```
TPolina, line 15: you don't use key here, so iterate through just values

main.py
```diff
@@ -1,10 +1,75 @@
...
+
+
+def create_race_and_skill_model():
+    """
+    This function making Skill and Race models
+    """
+    data = get_race_info()
+    for race in data:
+        if not Race.objects.filter(name=race["name"]).exists():
+            Race.objects.create(name=f"{race['name']}",
```
TPolina, line 25: `race['name']` is already a string, no need to use `f-string`

main.py
```diff
@@ -1,10 +1,75 @@
...
+                                description=f"{race['description']}")
+
+        for skill in race["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(name=skill['name'],
```
TPolina, line 30: use double quotes everywhere

main.py
```diff
@@ -1,10 +1,75 @@
...
+                                     bonus=skill['bonus'],
+                                     race=Race.objects.get(
+                                         name=f"{race['name']}"))
```
TPolina, line 33: better get `race` once before iterating through `skills` instead of making a query for each `skill`

main.py
```diff
@@ -1,10 +1,75 @@
...
+
+
+def create_guild_model():
+    """
+    This function making Guild models
+    """
+    data = load_data_json()
+    guilds_data = [value["guild"] for key, value in data.items()]
```
TPolina, line 41: key is not used

main.py
```diff
@@ -1,10 +1,75 @@
...
+    for guild in guilds_data:
+        if guild is not None:
+            if not Guild.objects.filter(name=guild["name"]).exists():
+                if guild["description"] is not None:
```
TPolina, line 45: you can omit this check. When you create a `guild` description will be set to some value or None

TPolina (changes requested):

main.py
BohdanKryven: Thanks! I fixed all problems

TPolina (approved):

PR #41, opened 2022-06-20
Solution

Magus6784:


main.py
```diff
@@ -1,10 +1,58 @@
 import init_django_orm  # noqa: F401
 
+import json
```
Y-Havryliv, line 3: wrong import order(json should be imported first)

Y-Havryliv (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
```
Y-Havryliv, line 16: also guild should not be unique

main.py
Magus6784: Why?

lenasharabura (approved):

main.py
Magus6784: Your link does not want to open, send in telegram

TPolina (approved):
Good solution!

PR #49, opened 2022-06-20
Solution

RomaTsybulia:


main.py
```diff
@@ -1,10 +1,98 @@
...
 from db.models import Race, Skill, Player, Guild
 
+BASE_DIR = Path(__file__).resolve().parent
+
+
+def create_guild(guild: dict):
+    if guild:
+        guild_name = guild["name"]
+        if not Guild.objects.filter(name=guild_name).exists():
+            if guild["description"]:
```
katevasilenko, line 15: I guess this check is not necessary

katevasilenko (approved):

db/models.py
```diff
@@ -1 +1,43 @@
...
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE
+    )
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True
+    )
```
TPolina, line 42: this field cannot be null or blank, because the player will be deleted when Guild is deleted. You should set another `on_delete` value, so that player will be not deleted. `blank=True` is unnecessary in any cases

main.py
TPolina: yes, you can just create a guild and the description will be set to null automatically

main.py
```diff
@@ -1,10 +1,98 @@
...
+                guild_description = guild["description"]
+                Guild.objects.create(
+                    name=guild_name,
+                    description=guild_description
+                )
+            else:
+                Guild.objects.create(name=guild_name)
```
TPolina, line 22: this block is not needed

db/models.py
```diff
@@ -1 +1,43 @@
 from django.db import models
+
+
+class Race(models.Model):
+    races = [
```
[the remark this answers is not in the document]
TPolina: agree, you don't need choices here

main.py
```diff
@@ -1,10 +1,98 @@
...
+            description=race_description
+        )
+    return race_name
+
+
+def create_skills(skills: dict, race_name: str):
+    race_skills = {
+        skill["name"]: skill["bonus"]
+        for skill in skills
+    }
```
TPolina, line 40: You create a dict with incorrect logic: why "name" is a key and "bonus" is a value? That's not a good practice. `skills` is a list of correct dictionaries. Just iterate through each skill and use skill["name"], skill["bonus"].

main.py
```diff
@@ -1,10 +1,98 @@
...
+    return None
+
+
+def create_race(race_name: str, race_description: str):
+    if not Race.objects.filter(name=race_name).exists():
+        Race.objects.create(
+            name=race_name,
+            description=race_description
+        )
+    return race_name
```
TPolina, line 33: better return `Race` object here, not just `race_name`. Then you can use this object and don't make too many queries to the database

main.py
```diff
@@ -1,10 +1,98 @@
...
+        if not Guild.objects.filter(name=guild_name).exists():
+            if guild["description"]:
+                guild_description = guild["description"]
+                Guild.objects.create(
+                    name=guild_name,
+                    description=guild_description
+                )
+            else:
+                Guild.objects.create(name=guild_name)
+        return guild_name
```
TPolina, line 23: better return the `Guild` object here, not just `guild_name`. Then you can use this object and simplify your code

main.py
```diff
@@ -1,10 +1,98 @@
...
+                race=Race.objects.get(name=race),
+                guild=Guild.objects.get(name=guild_name)
+            )
+        else:
+            Player.objects.create(
+                nickname=nickname,
+                email=email,
+                bio=bio,
+                race=Race.objects.get(name=race)
+            )
```
TPolina, line 72: if you return the `Guild` object, not `guild_name` in 'create_guild', you will not need this `else` block

main.py
```diff
@@ -1,10 +1,98 @@
...
+    race_skills = {
+        skill["name"]: skill["bonus"]
+        for skill in skills
+    }
+    for key, value in race_skills.items():
+        if not Skill.objects.filter(name=key).exists():
+            Skill.objects.create(
+                name=key,
+                bonus=value,
+                race=Race.objects.get(name=race_name)
```
TPolina, line 46: if you return the `Race` object, not `race_name` in 'create_race', you will be able to use this object here

main.py
```diff
@@ -1,10 +1,98 @@
...
+                nickname=nickname,
+                email=email,
+                bio=bio,
+                race=Race.objects.get(name=race)
+            )
+
 
 def main():
-    pass
+    with open(BASE_DIR / "players.json", "r") as data:
```
TPolina, line 76: you don't need a path, you can open file with simple `with open("players.json", "r")`

TPolina (changes requested):

TPolina (approved):

PR #59, opened 2022-06-29
Develop1

Yurnerroo:


Aleksei-Isaev (approved):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
TPolina, line 25: this field cannot be `null`, because the player will be deleted when Guild is deleted. You should set another `on_delete` value, so that player will be not deleted.

TPolina (changes requested):

MasterpieceElbow (approved):

PR #73, opened 2022-08-07
solution

tarkaiev:


db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.CharField(max_length=255)
```
Abnormaltype, line 22: should be `emailField`

Abnormaltype (changes requested):

db/models.py
tarkaiev: updated

Abnormaltype (approved):
Well done!

PR #87, opened 2022-09-21
Solution

InnaKushnir:


IvanRamyk (changes requested):
add migrations to the PR to pass tests

PR #90, opened 2022-09-24
Task Complete

Glasis9:


Glasis9:
All tests pass in my IDE (if I comment out the line "addopts = --reuse-dp" in "pytest.ini")
If this is not done, then pytest does not work, I wrote about it in slake

main.py
```diff
@@ -1,10 +1,72 @@
...
+            guild = Guild.objects.get(
+                name=f'{players[player]["guild"]["name"]}')
+        else:
+            guild = None
+
+        Player.objects.create(
+            nickname=f'{player}',
+            email=f'{players[player]["email"]}',
+            bio=f'{players[player]["bio"]}',
+            race=Race.objects.get(name=f'{players[player]["race"]["name"]}'),
```
DmytroSvirsa, line 67: it must be a string already so you don t need f strings

DmytroSvirsa (changes requested):
make it with one loop `for a player in players`, now it not efficient code

main.py
DmytroSvirsa: also you now the data you got, don't use try/except

Glasis9: Now I am reworking the try-except constructs and will fix these errors

DmytroSvirsa (approved):

PR #102, opened 2022-10-05
Solution

wQuelS:


AriohBelskij (approved):
Good!

db/models.py
```diff
@@ -1 +1,32 @@
...
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race, on_delete=models.CASCADE
+    )
+    guild = models.ForeignKey(
+        Guild, on_delete=models.CASCADE, null=True
+    )
+    created_at = models.DateField(auto_now=True)
```
Nattalli, line 32: 1) should be `DateTimeField`
2) Also, `auto_now_add` would be better here. As you had to do this: `a DateTime field, that is set with the current time by default`. The difference is:
- when you use `auto_now_add` - records the time and date of creation
- when you use `auto_now` - updates the value of the field to the current time and date every time when the object changes

main.py
```diff
@@ -1,10 +1,47 @@
...
 
 from db.models import Race, Skill, Player, Guild
+import json
 
 
 def main():
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+        for player, info in data.items():
```
Nattalli, line 10: Should all this code be executed in this `with ... open ` block?

main.py
```diff
@@ -1,10 +1,47 @@
...
+            if info["guild"] is None:
+                guild = None
+            else:
+                if not Guild.objects.filter(
+                        name=info["guild"]["name"]
+                ).exists():
+                    guild = Guild.objects.create(
+                        name=info["guild"]["name"],
+                        description=info["guild"]["description"]
+                    )
```
Nattalli, line 35: what the value would be in the `guild` variable if 
```python 
info["guild"] is not None
``` 
and it already exists?

Nattalli (changes requested):

main.py
```diff
@@ -4,45 +4,43 @@
...
+        for skill in info["race"]["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race
                 )
 
+        if info["guild"]:
+            guild = Guild.objects.get_or_create(
```
Nattalli, line 28: wow, that looks very good!

Nattalli (approved):
You made a good job!

PR #104, opened 2022-10-06
Solution

AriohBelskij:


wQuelS (approved):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
Nattalli, line 26: Don't you think that `auto_now_add` would be better here? As you had to do this: `a DateTime field, that is set with the current time by default`. The difference is:
- when you use `auto_now_add` - records the time and date of creation
- when you use `auto_now` - updates the value of the field to the current time and date every time when the object changes

Nattalli (changes requested):
Good job. 1) resolve conflicts 2) check one left comment

Nattalli (approved):
Well done!

Nattalli (approved):

PR #105, opened 2022-10-06
Solution

sbublinskyi:


main.py
```diff
@@ -1,10 +1,75 @@
...
+        )
+
+
+def create_player(
+        player_name: str,
+        player: dict,
+        race: int,
+        guild: int,
+) -> None:
+    if len(Player.objects.filter(nickname=player_name)) == 0:
```
mltdll, line 35: You may simplify this check using `Model.objects.filter(some_field='data').exists()` method. But I'm not sure if this is an issue.

mltdll (changes requested):
Looks good to me, except for one thing.

mltdll (approved):

mltdll (changes requested):
Oops, conflicts.

mltdll (approved):

db/models.py
```diff
@@ -1 +1,54 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(
+        max_length=255,
+        unique=True
+    )
```
TPolina, line 8: In my opinion, it's better to write everything in one line where it's possible. If a line is too long then of course you need to split it into a few lines

db/models.py
```diff
@@ -1 +1,54 @@
...
+    bonus = models.CharField(
+        max_length=255,
+    )
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.PROTECT
```
TPolina, line 22: PROTECT is not the best option here. The race can be deleted and skills should be deleted as well

main.py
```diff
@@ -1,10 +1,70 @@
...
+            guild = create_guild(data["guild"])
+
+        create_player(name, data, race, guild)
+
+
+def create_player(
+        player_name: str,
+        player: dict,
+        race: int,
+        guild: int,
```
TPolina, line 28: better name them race_id and guild_id because now it's a bit confusing

TPolina (changes requested):

db/models.py
TPolina: You had PROTECT in the Player model and it was perfect. In that case you couldn't delete race when players exist. But when there are no players you can delete race and skills should be deleted as well. Now you changed race to CASCADE in Player model too. It's also possible depending on your logic but PROTECT there was better.

TPolina (approved):
Good job!

db/models.py
sbublinskyi: Than

> You had PROTECT in the Player model and it was perfect. In that case you couldn't delete race when players exist. But when there are no players you can delete race and skills should be deleted as well. Now you changed race to CASCADE in Player model too. It's also possible depending on your logic but PROTECT there was better.

Thanks)

PR #108, opened 2022-10-07
Solution

StasOkhrym:


main.py
```diff
@@ -1,10 +1,59 @@
...
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+        for player_nickname, player_data in data.items():
```
mltdll, line 12: I'm not if it's an issue, but I would exit the `with` block as soon as I finished the last operation with file. No need to keep useless references in memory, right?

main.py
```diff
@@ -1,10 +1,59 @@
...
+            race = create_race(player_data)
+            guild = create_guild(player_data)
+            create_skills(player_data)
+
+            if not Player.objects.filter(nickname=player_nickname).exists():
```
mltdll, line 17: In this task, there's no need to check if a player already exists. It doesn't hurt, but it's certainly not necessary.

mltdll (changes requested):

mltdll (approved):

Nattalli (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
+    created_at = models.DateField(auto_now=True)
```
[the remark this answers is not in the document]
StasOkhrym: Got it, thank you 🙂

PR #109, opened 2022-10-08
'Solution'

davidkrivko:


IvanRamyk (changes requested):
tests aren't passed

davidkrivko:
I don't understand what is wrong

main.py
```diff
@@ -1,10 +1,55 @@
...
+        for key, value in info.items():
+            if key == "race":
+                if not Race.objects.filter(name=value["name"]).exists():
+                    Race.objects.create(
+                        name=value["name"],
+                        description=value["description"],
+                    )
+                if value["skills"] is not None:
+                    for skill in value["skills"]:
+                        i_ = Skill.objects.filter(name=skill["name"]).exists()
```
IvanRamyk, line 20: `i_` isn't the best variable name :)

main.py
```diff
@@ -1,10 +1,55 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file_in:
+        data = json.load(file_in)
+    for player, info in data.items():
+        for key, value in info.items():
```
IvanRamyk, line 11: it seems that this loop is extra
you can simply check `info["race"]` and `info["guild"]` then

main.py
```diff
@@ -1,10 +1,55 @@
...
+        if player["guild"] is None:
+            Player.objects.create(
+                nickname=name,
+                email=player["email"],
+                bio=player["bio"],
+                race_id=Race.objects.get(name=player["race"]["name"]).id,
+                guild_id=None,
+            )
+        else:
+            Player.objects.create(
```
IvanRamyk, line 46: it's a duplication
just use a ternary operator to define `guild_id`

IvanRamyk (changes requested):

main.py
IvanRamyk: still not fixed

main.py
IvanRamyk: still not fixed

IvanRamyk (changes requested):

main.py
```diff
@@ -1,10 +1,51 @@
...
+            bio=player["bio"],
+            race_id=Race.objects.get(name=player["race"]["name"]).id,
+            guild_id=None,
+        ) if player["guild"] is None else Player.objects.create(
+            nickname=name,
+            email=player["email"],
+            bio=player["bio"],
+            race_id=Race.objects.get(name=player["race"]["name"]).id,
+            guild_id=Guild.objects.get(name=player["guild"]["name"]).id,
+        )
```
DmytroSvirsa, line 48: it's a duplication
just use a ternary operator to define guild_id

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #110, opened 2022-10-10
Solution without 'get_or_create' method

IvanStored:


wQuelS (approved):
GJ

db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
```
TPolina, line 6: it can be blank

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
TPolina, line 25: player should not be deleted when guild is deleted

TPolina (changes requested):

TPolina (approved):
Good job!

PR #115, opened 2022-10-14
solution

KirillDrago:


main.py
```diff
@@ -1,10 +1,52 @@
...
+        race = Race.objects.get(name=races["name"])
+
+        for skill in skills:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race
+                )
+        if guilds:
```
mltdll, line 32: I'd suggest to remove outer `if ... else` here and just assign the output of `create()` to `guilds` in the inner `if` block. By the way, why not name it `guild` right away?

mltdll (changes requested):

main.py
KirillDrago: Thanks for ur advice, but i can't implement it correctly without errors. So let's wait for mentor review. ;)
 (`guild` name is fixed)

mltdll: At this point you can simply remove the `else` instruction. Since an empty `giuld` field is represented as `null` in json, it's already `None` if it's empty.

mltdll (approved):

db/models.py
```diff
@@ -1 +1,32 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "Short description provided by a user about himself/herself",
```
TPolina, line 27: same here: the description is unnecessary.

db/models.py
```diff
@@ -1 +1,32 @@
...
+        max_length=255
+    )
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
TPolina, line 31: player should not be deleted when guild is deleted

TPolina (changes requested):

TPolina (approved):
Good job!

PR #116, opened 2022-10-16
Solution

luxuriant777:


main.py
```diff
@@ -1,10 +1,46 @@
...
+            for skill in race_dict["skills"]:
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race)
+        else:
+            race = Race.objects.get(name=race_dict["name"])
+
+        guild_dict = player_info["guild"]
+        if guild_dict is not None:
```
Gr1zzZzzly, line 27: i think u can use just 
if guild_dict:
....
...

Gr1zzZzzly (changes requested):

main.py
luxuriant777: Thanks. Changed

PR #121, opened 2022-10-18
Solution

taras-andruschenko:


db/models.py
```diff
@@ -1 +1,34 @@
...
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(
+        "This field describes what kind of bonus "
+        "players can get from it",
```
TPolina, line 13: the description is unnecessary. Field name is enough to understand what it is

db/models.py
```diff
@@ -1 +1,34 @@
...
+        max_length=255
+    )
+    race = models.ForeignKey(Race, on_delete=models.DO_NOTHING)
```
TPolina, line 16: when the race is deleted skills should be deleted as well

db/models.py
```diff
@@ -1 +1,34 @@
...
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "It stores a short description "
+        "provided by a user about himself/herself",
+        max_length=255
+    )
+    race = models.ForeignKey(Race, on_delete=models.DO_NOTHING)
```
TPolina, line 32: when the race is deleted players should be deleted as well

db/models.py
```diff
@@ -1 +1,34 @@
...
+    guild = models.ForeignKey(Guild, on_delete=models.DO_NOTHING, null=True)
```
TPolina, line 33: when the guild is deleted it should become Null for the player. `Do_nothing` is a bad option

db/models.py
```diff
@@ -1 +1,34 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "It stores a short description "
+        "provided by a user about himself/herself",
```
TPolina, line 29: description in unnecessary

TPolina (changes requested):

TPolina (approved):
Good job!

PR #125, opened 2022-10-19
Solution

serviza:


db/models.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True, default=None)
```
Y-Havryliv, line 22: `default=None` is a bad option for guild name which is unique

db/models.py
```diff
@@ -1 +1,32 @@
...
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
Y-Havryliv, line 31: player should not be deleted when the guild is deleted

main.py
```diff
@@ -1,10 +1,50 @@
 import init_django_orm  # noqa: F401
-
+import json
```
Y-Havryliv, line 2: optimize your imports

Y-Havryliv (changes requested):

db/models.py
serviza: Agree.

db/models.py
serviza: Checked

main.py
serviza: I don't understand what I have to do.

Nattalli: @serviza Change the order of your imports according to `pep8`. Here `json` should be first.

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```

main.py
```diff
@@ -1,10 +1,50 @@
...
+
+        for skill in fields["race"]["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                skill = Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race
+                )
+            else:
+                skill = Skill.objects.get(name=skill["name"])
```
Nattalli, line 27: that is unnecessary

main.py
```diff
@@ -1,10 +1,50 @@
...
+                guild = Guild.objects.create(
+                    name=fields["guild"]["name"],
+                    description=fields["guild"]["description"]
+                )
+            else:
+                guild = Guild.objects.get(name=fields["guild"]["name"])
+        else:
+            guild = None
+
+        if not Player.objects.filter(nickname=player).exists():
```
Nattalli, line 40: and this check is also unnecessary

Nattalli (changes requested):

db/models.py
Sorry...

main.py
serviza: Thanks for explanation)

Nattalli (approved):
👍🏻

PR #128, opened 2022-10-19
Solution

shurushku:


main.py
```diff
@@ -1,10 +1,42 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
+def get_race(dict_race: dict) -> Race:
+    race = Race.objects.update_or_create(
```
mltdll, line 9: I don't think `update_or_create` is what you want to do here. Have you considered `get_or_create()`?

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Well done!

PR #129, opened 2022-10-19
Created models with migrations with diff fields params

alexzubkoff:


VadimSabadyr (approved):
Looks good for me!

db/models.py
```diff
@@ -1 +1,35 @@
...
+    description = models.TextField(blank=True)
+
+    def __str__(self) -> None:
+        return "%s %s" % (self.name, self.description)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE, null=True)
```
Y-Havryliv, line 15: `race` can't be null here

db/models.py
```diff
@@ -1 +1,35 @@
...
+    def __str__(self) -> None:
+        return "%s %s" % (self.name, self.description)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE, null=True)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
Y-Havryliv, line 34: player should not be deleted when the guild is deleted

db/models.py
```diff
@@ -1 +1,35 @@
...
+    created_at = models.DateTimeField(auto_now=True)
```
Y-Havryliv, line 35: `auto_now_add` would be better here

db/models.py
```diff
@@ -1 +1,35 @@
...
+
+    def __str__(self) -> None:
+        return "%s %s" % (self.name, self.description)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE, null=True)
```
Y-Havryliv, line 33: `race` can't be null

main.py
```diff
@@ -1,10 +1,52 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json") as json_file:
+        players_data = json.load(json_file)
+
+    for nickname, player_data in players_data.items():
+        if player_data["race"]:
```
Y-Havryliv, line 12: each player should have a race, so this check is not needed

main.py
```diff
@@ -1,10 +1,52 @@
...
+            guild = None
+
+        if not Player.objects.filter(nickname=nickname).exists():
+            Player.objects.create(
+                nickname=nickname,
+                email=player_data["email"],
+                bio=player_data["bio"],
+                race=race,
+                guild=guild,
+                created_at=datetime.now()
```
Y-Havryliv, line 48: `created_at` will be added automatically, because you have `auto_add_now`

Y-Havryliv (changes requested):

db/models.py
```diff
@@ -1 +1,36 @@
 from django.db import models
+from django.db.models import SET_NULL
```
Nattalli, line 2: is it really needed to import?

db/models.py
```diff
@@ -1 +1,36 @@
...
+        return "%s %s" % (self.name, self.description)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+    def __str__(self) -> None:
+        return "%s %s %s" % (self.name, self.bonus, self.race)
```
Nattalli, line 19: interesting. Why don't you use f-string here?

main.py
```diff
@@ -1,10 +1,49 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json") as json_file:
+        players_data = json.load(json_file)
+
+    for nickname, player_data in players_data.items():
+        if not Race.objects.filter(
+            name=player_data["race"]["name"]
+        ).exists():
```
Nattalli, line 13: unnecessary condition, you can remove it

main.py
```diff
@@ -1,10 +1,49 @@
...
+                    )
+        if player_data["guild"]:
+            if not Guild.objects.filter(
+                name=player_data["guild"]["name"]
+            ).exists():
+                Guild.objects.create(
+                    name=player_data["guild"]["name"],
+                    description=player_data["guild"]["description"],
+                )
+            guild = Guild.objects.get(name=player_data["guild"]["name"])
```
Nattalli, line 35: can you optimize it? At least, you can make `race = Race.objects.create()...`. 

main.py
```diff
@@ -1,10 +1,49 @@
...
+        else:
+            guild = None
+
+        if not Player.objects.filter(nickname=nickname).exists():
```
Nattalli, line 39: this condition is unnecessary, you can remove it

main.py
```diff
@@ -1,10 +1,49 @@
...
+
+    for nickname, player_data in players_data.items():
+        if not Race.objects.filter(
+            name=player_data["race"]["name"]
+        ).exists():
+            Race.objects.create(
+                name=player_data["race"]["name"],
+                description=player_data["race"]["description"],
+            )
+        race = Race.objects.get(name=player_data["race"]["name"])
```
Nattalli, line 18: optimize it

Nattalli (changes requested):

main.py
```diff
@@ -27,16 +26,15 @@ def main() -> None:
...
+                guild = Guild.objects.create(
                     name=player_data["guild"]["name"],
                     description=player_data["guild"]["description"],
                 )
-            guild = Guild.objects.get(name=player_data["guild"]["name"])
         else:
             guild = None
 
-        if not Player.objects.filter(nickname=nickname).exists():
+        if not Player.objects.filter(nickname=nickname):
```
Nattalli, line 37: good but you can remove the full condition: `if not Player.objects.filter`

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #132, opened 2022-10-19
Solution

Andriy-Makarenko:


db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
mltdll, line 24: I don't think it's a good idea to use cascade deletion on the user table. It seems to me that if a race was deleted by mistake, it's still better to be able to recover the players, because user data is the most valuable data in the database. At least, in this case.

mltdll (changes requested):

mltdll (approved):
I checked other solutions, and the mentors seem to approve solution like yours.

mltdll (approved):

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
TPolina, line 24: a player cannot be without race. Choose another on_delete option

TPolina (changes requested):

db/models.py
TPolina: Protect and cascade are ok. Set_null is not according to the task description

TPolina (approved):
Good job!

PR #133, opened 2022-10-19
Solution

VadymShkarbul:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 25: Since `guild` field can have a `null` value, cascade deletion should not be used.

main.py
```diff
@@ -1,10 +1,55 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+
+    with open("players.json", "r") as file:
+        data = json.load(file)
+
+    for players_name, other_info in data.items():
```
mltdll, line 12: I think it's better to use simple singular `player`, because now it looks like plural, even id you meant `player's` here. You may also just get rid of `players_` part in all of the variables. It doesn't make much difference here.

mltdll (changes requested):

db/models.py
VadymShkarbul: Thank you for the review. You are right and I have read again https://medium.com/@inem.patrick/django-database-integrity-foreignkey-on-delete-option-db7d160762e4. But I need to read a lot about git too. Someting went wrong, so I was forced to do new pull request https://github.com/mate-academy/py-game-models/pull/135 please check it

PR #135, opened 2022-10-20
Solution wiht fixed on_delete option

VadymShkarbul:


mltdll (approved):

main.py
```diff
@@ -1,10 +1,55 @@
 import init_django_orm  # noqa: F401
+import json
```
Y-Havryliv, line 2: optimize your imports

Y-Havryliv (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```

main.py
```diff
@@ -1,10 +1,56 @@
...
+        player_race = other_info["race"]
+        player_guild = other_info["guild"]
+        player_skills = player_race["skills"]
+
+        if not Race.objects.filter(name=player_race["name"]).exists():
+            Race.objects.create(
+                name=player_race["name"],
+                description=player_race["description"]
+            )
+        player_race = Race.objects.get(name=player_race["name"])
```
Nattalli, line 26: can you optimize it? At least, you can make `race = Race.objects.create()...`. 

main.py
```diff
@@ -1,10 +1,56 @@
...
+
+        if player_guild:
+            if not Guild.objects.filter(name=player_guild["name"]).exists():
+                Guild.objects.create(
+                    name=other_info["guild"]["name"],
+                    description=other_info["guild"]["description"]
+                )
+            player_guild = Guild.objects.get(name=player_guild["name"])
```
Nattalli, line 34: the same here

main.py
```diff
@@ -1,10 +1,56 @@
...
+
+        for skill in player_skills:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=player_race
+                )
+
+        if not Player.objects.filter(nickname=player_name).exists():
```
Nattalli, line 46: this check is unnecessary here

Nattalli (changes requested):

db/models.py
Sorry...

main.py
VadymShkarbul: Thanks! I've read about get_or_create after finishing the task, so I used it now! :)

Nattalli (approved):
Now it looks much better

PR #139, opened 2022-10-20
'Solution'

OvDima1:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 25: Since this field can be `null`, `CASCADE` option shouldn't be used here.

mltdll (changes requested):

mltdll (approved):
By the way, judging by your commit names, it seems that single quotes don't work in your terminal. Why don't you try double quotes? :)

OvDima1:
> By the way, judging by your commit names, it seems that single quotes don't work in your terminal. Why don't you try double quotes? :)

OvDima1:
> By the way, judging by your commit names, it seems that single quotes don't work in your terminal. Why don't you try double quotes? :)

OvDima1:
> By the way, judging by your commit names, it seems that single quotes don't work in your terminal. Why don't you try double quotes? :)
Never even paid attention to it :)

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```

main.py
```diff
@@ -2,9 +2,47 @@
 
 from db.models import Race, Skill, Player, Guild
 
+import json
+
```
Nattalli, line 6: Change the order of your imports according to `pep8`. Here `json` should be first.

main.py
```diff
@@ -2,9 +2,47 @@
...
+    with open("players.json", "r") as file:
+        data = json.load(file)
+
+    for nickname, info in data.items():
+        if not Race.objects.filter(name=info["race"]["name"]).exists():
+            race = Race.objects.create(
+                name=info["race"]["name"],
+                description=info["race"]["description"]
+            )
+        race = Race.objects.get(name=info["race"]["name"])
```
Nattalli, line 18: can you optimize it? At least, you can make `race = Race.objects.create()...`. 
Btw, It is best practice to use the `get _or _create ()` method in general, as this, will improve readability
That is, you MUST reduce the number of requests to the database, but not necessarily use `get_or_create()`(that is only best practice for now)

Nattalli (changes requested):

db/models.py
Nattalli: Sorry!

Nattalli (approved):
Well done!

PR #140, opened 2022-10-20
Solution

graandda:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 25: It's a bit harsh to delete all the players when the guild they were in is deleted, don't you think?

main.py
```diff
@@ -1,10 +1,63 @@
...
+
+            if not Race.objects.filter(
+                    name=player_data["race"]["name"]).exists():
+                player_race = Race.objects.create(
+                    name=player_data["race"]["name"],
+                    description=player_data["race"]["description"]
+                )
+            else:
+                player_race = Race.objects.get(
+                    name=player_data["race"]["name"])
```
mltdll, line 24: This formatting is not considered ideal here. If you start a new line after the opening parenthesis, you should also do the same before the closing one. Same at $17, $40, $47.

mltdll (changes requested):

db/models.py
graandda: Thank you)

main.py
graandda: yep!

mltdll: Just so you know: so far you fixed it only at line 47. :)

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```

main.py
```diff
@@ -1,10 +1,64 @@
 import init_django_orm  # noqa: F401
 
+import json
+
 from db.models import Race, Skill, Player, Guild
```
Nattalli, line 5: Change the order of your imports according to `pep8`. Here `json` should be first.

main.py
```diff
@@ -1,10 +1,64 @@
...
+            # RACE
+
+            if not Race.objects.filter(
+                    name=player_data["race"]["name"]).exists():
+                player_race = Race.objects.create(
+                    name=player_data["race"]["name"],
+                    description=player_data["race"]["description"]
+                )
+            else:
+                player_race = Race.objects.get(
```
Nattalli, line 23: read 4.1 here:
 https://github.com/mate-academy/py-game-models/blob/master/checklist.md

```
The context manager is needed to work with the file, in our case, to read data. Therefore, after you have read the data from the file, exit the block.
```

Nattalli (changes requested):

db/models.py
Sorry...

Nattalli (approved):
👍🏻

PR #141, opened 2022-10-20
Solve the task

Cerne13:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 25: I don't think `on_delete` parameter should be set to `CASCADE` for guild field here. This field can be null by design, so the table's integrity is not in danger if player's guild is deleted. :)

main.py
```diff
@@ -1,10 +1,44 @@
...
+        for skill in race["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=Race.objects.get(name=race["name"])
+                )
+
+        guild = player_data["guild"]
+        if guild and not Guild.objects.filter(name=guild["name"]).exists():
```
mltdll, line 29: I don't like the way you check the same `guild` value twice, even though I understand that the code looks better this way.

By the way, it would be more efficient in terms of database interaction to store the return value of `.create()` and use `.get()` only if you need to. But yet again, the code won't be as pretty as it is now.

The other option is to use `.get_or_create()`.

mltdll (changes requested):

db/models.py
Cerne13: Thank you! Yes, it's too harsh to delete a player if a guild is deleted :)

main.py
Cerne13: Much appreciated!
Thank you, especially for the `.get_or_create()` hint. With it the code will be much shorter and readable.
Guess I'll refactor in between today's kick-offs:

mltdll (approved):
Looks great!

main.py
```diff
@@ -1,10 +1,42 @@
...
+        skills = race["skills"]
+        guild = player_data["guild"]
+
+        race = Race.objects.get_or_create(
+            name=race["name"],
+            description=race["description"]
+        )[0]
+
+        for skill in skills:
+            Skill.objects.get_or_create(
```
TPolina, line 22: you don't need `get_or_create` here so use simple `create`

TPolina (changes requested):

main.py
Cerne13: It won't work this way. 
A skill model's name must be unique - so if we parse the json file with just `createI()` we'll get an error because of the doubling skills from different players (we have 3 players with same skillsets).

Cerne13: Instead implemented get_or_create() to player creation. 
(So the script won't fail in case we occasionally run the it several times or there is an existing user among new ones in a file)

MasterpieceElbow (approved):
GJ!

PR #142, opened 2022-10-20
Solution

Vasyl-Poremchuk:


LiudmylaKulinchenko (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
```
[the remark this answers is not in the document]
Vasyl-Poremchuk: Thanks for your review)

LiudmylaKulinchenko (approved):
Good job!

db/models.py
```diff
@@ -1 +1,32 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
Y-Havryliv, line 31: player should not be deleted when the guild is deleted

main.py
```diff
@@ -1,10 +1,52 @@
...
+            current_guild = None
+        else:
+            if not Guild.objects.filter(name=guild["name"]).exists():
+                Guild.objects.create(
+                    name=guild["name"],
+                    description=guild["description"],
+                )
+            current_guild = Guild.objects.get(name=guild["name"])
+
+        if not Player.objects.filter(nickname=player).exists():
```
Y-Havryliv, line 42: It' no need to check if a player already exists

db/models.py
```diff
@@ -1 +1,32 @@
...
+    created_at = models.DateTimeField(auto_now=True)
```
Y-Havryliv, line 32: `auto_now_add` would be a better option here

Y-Havryliv (changes requested):

db/models.py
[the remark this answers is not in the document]
Nattalli: @LiudmylaKulinchenko Btw, It is not required, solution without choices is correct as well

db/models.py
```diff
@@ -1 +1,32 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.PROTECT, null=True)
```
Nattalli, line 31: `models.SET_NULL` would be better here

Nattalli, line 31: add `related_name`, it's a best practice 😸

main.py
```diff
@@ -1,10 +1,51 @@
...
+        race = player_info["race"]
+        skills = race["skills"]
+        guild = player_info["guild"]
+
+        if not Race.objects.filter(name=race["name"]).exists():
+            Race.objects.create(
+                name=race["name"],
+                description=race["description"],
+            )
+        current_race = Race.objects.get(name=race["name"])
```
Nattalli, line 22: can you optimize it? At least, you can make `race = Race.objects.create()...`. 

main.py
```diff
@@ -1,10 +1,51 @@
...
+        if not guild:
+            current_guild = None
+        else:
+            if not Guild.objects.filter(name=guild["name"]).exists():
+                Guild.objects.create(
+                    name=guild["name"],
+                    description=guild["description"],
+                )
+            current_guild = Guild.objects.get(name=guild["name"])
+
```
Nattalli, line 41: the same here. you can optimize

Nattalli (changes requested):

db/models.py
Nattalli: Please do not add this. Yes, this is a best practice, but the tests are not ready for such code(
Sorry...

Nattalli (approved):
Well done!

PR #148, opened 2022-10-20
Add models

FranchukS:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
```
Vasyl-Poremchuk, line 5: You should add choices here

main.py
```diff
@@ -1,11 +1,49 @@
...
+    for nickname, info in players.items():
+
+        if info.get("guild"):
+
+            if not Guild.objects.filter(name=info["guild"]["name"]).exists():
+                guild = Guild.objects.create(
+                    name=info["guild"]["name"],
+                    description=info["guild"]["description"]
+                )
+            else:
```
Vasyl-Poremchuk, line 21: You don't need to use `else` here, you can just move this variable to the left

main.py
```diff
@@ -1,11 +1,49 @@
...
+        else:
+            guild = None
+
+        if not Race.objects.filter(name=info["race"]["name"]).exists():
+
+            race = Race.objects.create(
+                name=info["race"]["name"],
+                description=info["race"]["description"]
+            )
+        else:
```
Vasyl-Poremchuk, line 32: Same here

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
GJ!

MasterpieceElbow (approved):
GJ!

db/models.py
Vasyl-Poremchuk: Sorry for telling you to add `choices` here, I did as you did, but I got a comment about it, but it turns out it's not necessary)

PR #149, opened 2022-10-20
Solution

pavlo-kapitanets:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 25: I don't think `CASCADE` should be applied to this field. What happens if guild leader decides to delete the guild? Should all the players in it be deleted? :)

main.py
```diff
@@ -1,10 +1,50 @@
...
+                guild_id = Guild.objects.get(name=user_guild["name"])
+            else:
+                guild_id = None
+
+            for skill in user_skills:
+                if not Skill.objects.filter(name=skill["name"]).exists():
+                    Skill.objects.create(
+                        name=skill["name"],
+                        bonus=skill["bonus"],
+                        race=Race.objects.get(name=user_race["name"])
```
mltdll, line 38: Getting the race from the database for every skill in list seems terribly inefficient. Why don't you store it in a variable and reuse it?

main.py
```diff
@@ -1,10 +1,50 @@
...
+                    )
+
+            Player.objects.create(
+                nickname=user_name,
+                email=information["email"],
+                bio=information["bio"],
+                guild=guild_id,
+                race=Race.objects.get(name=user_race["name"])
```
mltdll, line 46: Don't forget to reuse it here as well.

mltdll (changes requested):
By the way, the same idea of avoiding repeating database queries can be applied elsewhere in your solution. Can you find where? It's optional, I guess.

main.py
```diff
@@ -19,6 +19,7 @@ def main() -> None:
                     name=user_race["name"],
                     description=user_race["description"]
                 )
+                race_id = Race.objects.get(name=user_race["name"])
```
mltdll, line 22: It's a `Race` object. I don't think it's justified to call the variable `_id`.

mltdll (changes requested):

db/models.py
pavlo-kapitanets: Thank's for your review! I've done almost all changes, but I don't know how to make my code more efficient.

mltdll: The solution I mean is not an improvement of Python code. It's an improvement of database interactions.

If you store the output of `.create()` methods in variables and only execute `.get()` when the entry already exists, it will result in less database calls overall.

There's also a method for this: `.get_or_create()`.

mltdll (changes requested):
Just making sure you've seen the reject I made before you requested review. :)

db/models.py
pavlo-kapitanets: I've tried to refactor my code with .get_or_create() method. I hope I took your point and did it right.

mltdll (approved):

db/models.py
mltdll: Absolutely, GJ! :)

main.py
```diff
@@ -1,10 +1,47 @@
...
+            user_skills = information["race"]["skills"]
+            user_guild = information["guild"]
+
+            race_of_user, _ = Race.objects.get_or_create(
+                name=user_race["name"],
+                description=user_race["description"]
+            )
+
+            if user_guild:
+                guild_id, _ = Guild.objects.get_or_create(
```
TPolina, line 23: it's not a `guild_id` but a Guild object so your variable name is confusing

TPolina (changes requested):
It's a good practice to squash migrations when you have too many of them that refer to the same models

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #150, opened 2022-10-21
Solution

misha-doroshenko:


db/models.py
```diff
@@ -1 +1,29 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 28: `CASCADE` might not be the best option here, since this field can be `null`.

main.py
```diff
@@ -1,10 +1,61 @@
...
+import json
 from db.models import Race, Skill, Player, Guild
+from django.db.models import QuerySet
+
+
+def create_guild(guild_info: None | dict) -> None | QuerySet:
+    if guild_info:
+        guild_name = guild_info["name"]
+        if not Guild.objects.filter(name=guild_name).exists():
+            Guild.objects.create(
```
mltdll, line 11: You can return the output of `.create()` method here.

main.py
```diff
@@ -1,10 +1,61 @@
...
+                description=guild_info["description"]
+            )
+        return Guild.objects.get(name=guild_name)
+    return None
+
+
+def create_race(race_info: dict) -> QuerySet:
+    race_name = race_info["name"]
+    if not Race.objects.filter(name=race_name).exists():
+        Race.objects.create(
```
mltdll, line 22: Same here.

mltdll (changes requested):

db/models.py
```diff
@@ -24,6 +24,6 @@ class Player(models.Model):
     nickname = models.CharField(max_length=255, unique=True)
     email = models.EmailField(max_length=255)
     bio = models.CharField(max_length=255)
-    race = models.ForeignKey(Race, on_delete=models.CASCADE)
-    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL)
```
mltdll, line 27: Race shouldn't actually have `SET_NULL` option. Only guild.

mltdll (changes requested):

db/models.py
misha-doroshenko: It depends on your realization. I don't want to lose all information about the player if the race will be deleted

mltdll: Then you should use `PROTECT`.

db/models.py
```diff
@@ -1 +1,29 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
+    guild = models.ForeignKey(Guild, on_delete=models.PROTECT, null=True)
```
mltdll, line 28: Look at it this way: a guild is an optional feature. A player can create a guild, delete it, enter, or leave. If a guild owner decides to delete his guild, should they:
- a) be able to delete the guild, in which case all the players in it will have no guild (`SET_NULL`),
- b) be able to delete the guild, in which case all the players in it will be deleted (`CASCADE`), or
- c) not be able to delete the guild as long as there's at least one player in it (`PROTECT`)?

mltdll (changes requested):

db/models.py
misha-doroshenko: ok, I understand

mltdll (approved):

db/models.py
```diff
@@ -1 +1,29 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(
+        "Describes what kind of bonus players can get from it",
```
TPolina, line 12: the description is unnecessary. Field name is enough to understand what it is

TPolina (approved):
Good job!

PR #152, opened 2022-10-21
Solution

pazyn404:


FranchukS (changes requested):
You need to add migrations to the repository, I guess it's the reason tests are failed

FranchukS (approved):
👍

db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(primary_key=True, max_length=255)
```
Nattalli, line 5: should be unique

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(primary_key=True, max_length=255)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
```
Nattalli, line 12: I think `CASCADE` would be better here

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(primary_key=True, max_length=255)
```
Nattalli, line 16: also unique

Nattalli: btw, `primary_key` as a string value is not a best practice. So you shouldn't change it

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(primary_key=True, max_length=255)
```
Nattalli, line 21: also unique

db/models.py
```diff
@@ -1 +1,26 @@
...
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```

main.py
```diff
@@ -1,10 +1,55 @@
+from json import load
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
     pass
```
Nattalli, line 9: Remove the unnecessary statement `pass`

Nattalli (changes requested):

db/models.py
Sorry...

Nattalli (approved):

PR #153, opened 2022-10-21
Solution

matechris:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.CASCADE, null=True)
```
mltdll, line 25: Since this field can be `null`, `CASCADE` option shouldn't be used here.

mltdll (changes requested):

mltdll (approved):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```

main.py
```diff
@@ -1,10 +1,61 @@
...
+
+
+def create_race(player_info: dict) -> Race:
+    if not Race.objects.filter(name=player_info["race"]["name"]).exists():
+        Race.objects.create(
+            name=player_info["race"]["name"],
+            description=player_info["race"]["description"]
+        )
+
+    return Race.objects.get(name=player_info["race"]["name"])
```
Nattalli, line 34: can you optimize it? At least, you can make `race = Race.objects.create()...`. 

Nattalli (changes requested):

db/models.py
Sorry...

Nattalli (approved):
Good job!

PR #159, opened 2022-10-24
game-model

VitaliyBychevoy:


main.py
```diff
@@ -1,10 +1,46 @@
+import json
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    file_name = "players.json"
```
Vasyl-Poremchuk, line 8: I think there is no need to create a variable and assign it a filename

main.py
```diff
@@ -1,10 +1,46 @@
...
+    with open(file_name) as file:
+        data = json.load(file)
+
+    for player_name, player_info in data.items():
+
```
Vasyl-Poremchuk, line 13: In my opinion, we don't need to add an empty line from which we start writing something in our code block)

Vasyl-Poremchuk (changes requested):

PR #160, opened 2022-10-24
Solution

YevheniiDavydenko:


db/models.py
```diff
@@ -1 +1,47 @@
-from django.db import models
+from django.db import models as m
```
mltdll, line 1: This alias does not look very good. I'd suggest to avoid using it unless it's a convention I'm not aware of.

db/models.py
```diff
@@ -1 +1,47 @@
...
+        ("D", "Dwarf"),
+        ("H", "Human"),
+        ("O", "Ork")
+    )
+
+    name = m.CharField(max_length=255, unique=True, choices=races)
+    description = m.TextField(blank=True)
+
+    def __str__(self) -> str:
+        return f"The race is:{self.name}"
```
mltdll, line 16: This string is missing a space after `:`.

db/models.py
```diff
@@ -1 +1,47 @@
...
+
+
+class Skill(m.Model):
+    name = m.CharField(max_length=255, unique=True)
+    bonus = m.CharField(max_length=255)
+    race = m.ForeignKey(Race, on_delete=m.PROTECT)
```
mltdll, line 22: `PROTECT` is not the best option here. Should you really be unable to delete a race without manually deleting all the skills?

db/models.py
```diff
@@ -1 +1,47 @@
...
+    def __str__(self) -> str:
+        return f"Skill: {self.name}, belongs to {self.race.name}"
+
+
+class Guild(m.Model):
+    name = m.CharField(max_length=255, unique=True)
+    description = m.TextField(blank=True, null=True)
+
+    def __str__(self) -> str:
+        return f"Guild {self.name}!"
```
mltdll, line 33: Looks like there should be a `:` here.

db/models.py
```diff
@@ -1 +1,47 @@
...
+
+
+class Player(m.Model):
+    nickname = m.CharField(max_length=255, unique=True)
+    email = m.EmailField(max_length=255)
+    bio = m.CharField(max_length=255)
+    race = m.ForeignKey(Race, on_delete=m.PROTECT)
+    guild = m.ForeignKey(Guild, on_delete=m.PROTECT, null=True)
```
mltdll, line 41: `guild` field also shouldn't be protected.

main.py
```diff
@@ -1,10 +1,53 @@
 import init_django_orm  # noqa: F401
-
+import json
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as f:
+        date = json.loads(f.read())
```
mltdll, line 8: You can use `json.load()` directly on `IO` object. That is, on `f`.

main.py
```diff
@@ -1,10 +1,53 @@
...
+
+    def create_race(date: dict) -> Race:
```
mltdll, line 10: It's not `date`, it's `data`. Please, fix it in all functions.

main.py
```diff
@@ -1,10 +1,53 @@
...
+        date = date.get("race")
```
mltdll, line 11: I don't really like the usage of `.get()` here, since it's possible to use `[]`, at least in most cases.
It applies to the entire file, not just this line.

main.py
```diff
@@ -1,10 +1,53 @@
...
+        if not Race.objects.filter(name=date.get("name")).exists():
+            Race.objects.create(
```
mltdll, line 13: You can `return` the output of `.create()` here: this way you'll avoid unnecessary `.get()`. Same in `create_guild`.

main.py
```diff
@@ -1,10 +1,53 @@
...
+            if not Guild.objects.filter(name=date.get("name")).exists():
+                Guild.objects.create(
+                    name=date.get("name"),
+                    description=date.get("description")
+                )
+            return Guild.objects.get(name=date.get("name"))
+        return None
+
+    for name, info in date.items():
+        race = create_race(info)
```
mltdll, line 40: Why pass the entire dictionary to these functions? Isn't it cleaner to pass only what's needed?

main.py
YevheniiDavydenko: It's my own prefer to use .get(), because you can indicate value when dict don't have some key, so I always use it.

mltdll (approved):

mltdll:
By the way, it's best practice to make your commit messages as easy to read and understand as possible. Especially the subject line.

MasterpieceElbow (approved):
GJ!

PR #161, opened 2022-10-24
test added

Y-Havryliv:


db/tests/test_main.py
```diff
@@ -72,3 +73,7 @@ def test_players():
 
 def test_email_field():
     assert isinstance(Player._meta.get_field("email"), EmailField)
+
+
+def test_guild_on_delete():
+    assert Player._meta.get_field("guild").remote_field.on_delete == models.SET_NULL
```
MasterpieceElbow: Maybe we should add the same on the foreign key `race`? Edit readme so the students will know what behavior to implement.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #162, opened 2022-10-24
Solution

ShevchukVitalii:


main.py
```diff
@@ -1,10 +1,50 @@
...
+            if not Guild.objects.filter(name=guild_data["name"]).exists():
+                Guild.objects.create(
+                    name=guild_data["name"],
+                    description=guild_data["description"]
+                )
+            guild = Guild.objects.get(name=guild_data["name"])
+        else:
+            guild = None
+
+        if not Player.objects.filter(nickname=player_nickname).exists():
```
Mowka7, line 40: this check is unnecessary

Mowka7 (changes requested):

main.py
```diff
@@ -1,10 +1,49 @@
...
+        players = json.load(players_file)
+
+    for player_nickname, player_data in players.items():
+        race_data = player_data["race"]
+        if not Race.objects.filter(name=race_data["name"]).exists():
+            Race.objects.create(
+                name=race_data["name"],
+                description=race_data["description"],
+            )
+        race = Race.objects.get(name=race_data["name"])
```
ArsenPidhoretskyi, line 19: In this case better to use [get_or_create](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create) 

The same fix with guild

ArsenPidhoretskyi (changes requested):

main.py
ShevchukVitalii: fixed

vsmutok (approved):
Looks Good 👍

PR #163, opened 2022-10-29
first solution

Foggy-Cat:


main.py
```diff
@@ -1,10 +1,56 @@
...
-    pass
+    with open("players.json", "r") as players:
+        all_players = json.load(players)
+
+    for player_name, player_info in all_players.items():
+        race = player_info["race"]
+        skills = race["skills"]
+        guild = player_info["guild"]
+
+        try:
```
IvanRamyk, line 17: I think this `try`-`except` block is difficult to read. Could you rewrite it using the `if`-`else` block?

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #166, opened 2022-11-03
'Solution'

IvanGLS:


main.py
```diff
@@ -1,10 +1,45 @@
 import init_django_orm  # noqa: F401
 
+import json
+
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as data_file:
```
IvanRamyk, line 9: it's hard to read this code. let's split it into several functions

IvanRamyk (changes requested):

main.py
IvanRamyk: let's create more functions :( 
it's still long

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #191, opened 2022-12-15
Done

DevSpaciX:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True, null=True)
```
YuriiKindrat, line 6: there is no task to make it nullable

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True, null=True)
+    description = models.TextField(blank=True, null=True)
```
YuriiKindrat, line 17: the same here

main.py
```diff
@@ -1,10 +1,42 @@
...
 def main() -> None:
-    pass
+    with open("players.json") as players:
+        players_data = json.load(players)
+    for person in players_data:
+        email_to_add = players_data[person]["email"]
+        bio_to_add = players_data[person]["bio"]
+        race_to_add = players_data[person]["race"]
+        guild_to_add = players_data[person]["guild"]
+        skill_to_add = players_data[person]["race"]["skills"]
```
YuriiKindrat, line 14: U can make here for person, info in players_data:
and then it will more comfortable to make variables

YuriiKindrat (changes requested):

OleksandrPozden (approved):

PR #199, opened 2022-12-29
Solution

darynka-h:


main.py
```diff
@@ -1,11 +1,43 @@
...
+                    description=info["guild"]["description"]
+                )
+            else:
+                guild = Guild.objects.get(name=info["guild"]["name"])
+            Player.objects.create(nickname=name, email=info["email"],
+                                  bio=info["bio"], race=race, guild=guild)
+
+        else:
+            Player.objects.create(nickname=name, email=info["email"],
+                                  bio=info["bio"], race=race)
```
YuriiKindrat, line 39: there is no need to make an else with 2 variants of player creation, u can make variable guild at the start and make it None - than at the end u can just add a guild to the player
Also u can use get_or_create method for guild here - just to write less code)

main.py
```diff
@@ -1,11 +1,43 @@
...
 
 
 if __name__ == "__main__":
-    main()
+    print(main())
```
YuriiKindrat, line 43: no need to use print here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #209, opened 2023-01-19
'Solution'

illiariaboy94:


main.py
```diff
@@ -1,10 +1,48 @@
...
+    with open("players.json", "r") as file:
+        data = json.load(file)
+
+    for player_nickname, info in data.items():
+        if not Race.objects.filter(name=info["race"]["name"]).exists():
+            Race.objects.create(
+                name=info["race"]["name"],
+                description=info["race"]["description"]
+            )
+        race = Race.objects.get(name=info["race"]["name"])
```
ArsenPidhoretskyi, line 18: if you create an object you do 3 query
I recommend you use Model.objects.get_or_create method

main.py
```diff
@@ -1,10 +1,48 @@
...
+
+        for skill in info["race"]["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race
+                )
```
ArsenPidhoretskyi, line 26: Use bulk_create instead it will call INSERT sql only once.
1. Generate objects in list
2. Use bulk_create(list)

main.py
```diff
@@ -1,10 +1,48 @@
...
+
+        if info["guild"]:
+            if not Guild.objects.filter(name=info["guild"]["name"]).exists():
+                Guild.objects.create(
+                    name=info["guild"]["name"],
+                    description=info["guild"]["description"]
+                )
+            guild = Guild.objects.get(name=info["guild"]["name"])
```
ArsenPidhoretskyi, line 34: Use get_or_create

main.py
```diff
@@ -1,10 +1,48 @@
...
+            guild = None
+
+        if not Player.objects.filter(nickname=player_nickname).exists():
+            Player.objects.create(
+                nickname=player_nickname,
+                email=info["email"],
+                bio=info["bio"],
+                race=race,
+                guild=guild
+            )
```
ArsenPidhoretskyi, line 45: Use get_or_create here

ArsenPidhoretskyi (changes requested):

main.py
illiariaboy94: Ok, good advice.

main.py
illiariaboy94: Ok.

illiariaboy94:
Ok

PR #210, opened 2023-01-20
game-models solution

AnnaKabatova:


main.py
```diff
@@ -1,10 +1,61 @@
...
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race
+                )
+            else:
+                Skill.objects.get(name=skill["name"])
+
+        if guild_dict is None:
+            guild = None
+        else:
```
YuriiKindrat, line 38: unnecessary else here, u can implement guild = None, and then check if guild_dict is not None
Also u can use __get_or_create__ method to optimize ur code

main.py
```diff
@@ -1,10 +1,61 @@
...
+        if not Player.objects.filter(nickname=player_name).exists():
+            Player.objects.create(
+                nickname=player_name,
+                email=player_info["email"],
+                bio=player_info["bio"],
+                race=race,
+                guild=guild,
+            )
+        else:
+            Player.objects.get(nickname=player_name)
```
YuriiKindrat, line 58: no need of this block of code, u are not using this player anywhere

YuriiKindrat (changes requested):

AnnaKabatova:
didn't fully understand how to work with the method "get_or_create", it was not considered in theory, but made other requested changes

YuriiKindrat (approved):

PR #220, opened 2023-02-01
Solution

YuriiVataschuk:


main.py
```diff
@@ -1,10 +1,44 @@
 import init_django_orm  # noqa: F401
+import json
```
Grayder0152, line 2: Your imports must be ordered.

main.py
```diff
@@ -1,10 +1,44 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file_in:
+        players_data = json.load(file_in)
+
+    for player, data in players_data.items():
+
+        if not Race.objects.filter(name=data["race"]["name"]).exists():
+            Race.objects.create(
```
Grayder0152, line 14: Please, save this to variable to avoid getting  record in 24 line.

main.py
```diff
@@ -1,10 +1,44 @@
...
+                name=data["race"]["name"],
+                description=data["race"]["description"]
+            )
+
+        for skill_info in data["race"]["skills"]:
+            if not Skill.objects.filter(name=skill_info["name"]).exists():
+                Skill.objects.create(
```
Grayder0152, line 21: same as above

main.py
```diff
@@ -1,10 +1,44 @@
...
+                    name=skill_info["name"],
+                    bonus=skill_info["bonus"],
+                    race=Race.objects.get(name=data["race"]["name"])
+                )
+
+        if data["guild"]:
+            if not Guild.objects.filter(name=data["guild"]["name"]).exists():
+                Guild.objects.create(
```
Grayder0152, line 29: same as above

Grayder0152 (changes requested):

main.py
YuriiVataschuk: image

in this case, it is used only once. Does it make sense to save it into a variable? And if so how to do it coorectly in this case?

main.py
YuriiVataschuk: Can you explain please?

Grayder0152: image
https://peps.python.org/pep-0008/

main.py
Grayder0152: Okey, you can remove `skill_name`.
Save object to variable only after creation, for example 
image
And then use it variable on 27, 42 and 43 lines

Grayder0152 (changes requested):

Vasyl-Smutok (changes requested):
please don't change tests

image

PR #224, opened 2023-02-05
'Solution'

Oleksiy-Liubchenko:


main.py
```diff
@@ -1,10 +1,52 @@
...
+
+    for player_name, info in players.items():
+        if not Race.objects.filter(
+                name=info["race"]["name"]).exists():
+            Race.objects.create(
+                name=info["race"]["name"],
+                description=info["race"]["description"]
+            )
+        race = Race.objects.get(
+            name=info["race"]["name"])
```
ArsenPidhoretskyi, line 19: Use can use get_or_create for this also)

main.py
```diff
@@ -1,10 +1,52 @@
...
+
+        if info["guild"] is not None:
+            if not Guild.objects.filter(
+                    name=info["guild"]["name"]).exists():
+                guild = Guild.objects.create(
+                    name=info["guild"]["name"],
+                    description=info["guild"]["description"]
+                )
+            guild = Guild.objects.get(
+                name=info["guild"]["name"])
```
ArsenPidhoretskyi, line 29: The same with https://docs.djangoproject.com/en/4.1/ref/models/querysets/#get-or-create

main.py
```diff
@@ -1,10 +1,52 @@
...
+        else:
+            guild = None
```
ArsenPidhoretskyi, line 31: Move it before if and not use else)

main.py
```diff
@@ -1,10 +1,52 @@
...
+        Player.objects.create(
+            nickname=player_name,
+            email=info["email"],
+            bio=info["bio"],
+            race=race,
+            guild=guild
+        )
+        skills = info["race"]["skills"] if info["race"]["skills"] else None
+        if skills:
+            for skill in skills:
```
ArsenPidhoretskyi, line 42: Why you need so much checks?
You can just do `for skill in info["race"]["skills"]:` if list is empty it not enter loop

ArsenPidhoretskyi (changes requested):

main.py
Oleksiy-Liubchenko: It will not pass one test if I do like this without else: 
        guild = None
        if info["guild"] is not None:
            if not Guild.objects.filter(
                    name=info["guild"]["name"]).exists():
                guild, created = Guild.objects.get_or_create(
                    name=info["guild"]["name"],
                    description=info["guild"]["description"]
                )

Oleksiy-Liubchenko: @ArsenPidhoretskyi also I have some questions: 
1) In models Skill and Player we have the same pieces of code (FK) "race = models.ForeignKey(Race, on_delete=models.CASCADE)". Can we write it like a link? For e.g. in Player model write "race = Skill.race"? 
2) Before pushing, I needed to delete db (checklist rule). But when I want to see how it works I'd like to have db again, but when I run the script without db, I have errors. To avoid this situation, I just need to do db manual (create file, than create field e.t.c.) or we have another quicker option?

ArsenPidhoretskyi: Why it not pass the test? It is so strange because the code will be the same) But okay)

ArsenPidhoretskyi: 1. No, because it is a little different. But, as for me, it is also strange)
Race has multiple skills. And race has multiple players

ArsenPidhoretskyi: 2. No need to delete db, just not commit it. You can add db to gitignore if you need. Use python manage.py migrate it will auto create

main.py
```diff
@@ -11,22 +11,18 @@ def main() -> None:
     for player_name, info in players.items():
         if not Race.objects.filter(
                 name=info["race"]["name"]).exists():
```
ArsenPidhoretskyi, line 13: If you use get_or_create no need .exists. It will improve code efficiency. Check in all places

ArsenPidhoretskyi (changes requested):

main.py
Oleksiy-Liubchenko: image
here scrin with this test

Oleksiy-Liubchenko: Probably the problem was in .exist method row. Without this, it will work normal

ArsenPidhoretskyi (approved):
Well done!

PR #226, opened 2023-02-06
Solution

ArtMot99:


main.py
```diff
@@ -1,10 +1,49 @@
...
 
 def main() -> None:
-    pass
+    # Open and load JSON file to "players" variable
+    with open("players.json") as players_file:
+        players = json.load(players_file)
+        for player in players:
+            race = players[player]["race"]
+            skills = players[player]["race"]["skills"]
+            guild = players[player]["guild"]
```
ArsenPidhoretskyi, line 13: Use .items() or .values() for dict. Getting in the loop by key the same dict like: `players[player]` is so bad

main.py
```diff
@@ -1,10 +1,49 @@
...
+            # Create Race objects if records not exist yet
+            if not Race.objects.filter(name=race["name"]).exists():
+                Race.objects.create(
+                    name=race["name"],
+                    description=race["description"]
+                )
```
ArsenPidhoretskyi, line 19: You can use .get_or_create
The same in other places

main.py
```diff
@@ -1,10 +1,49 @@
...
+            # Create Guild objects if records not exist yet
+            if guild:
```
ArsenPidhoretskyi, line 21: Can you explain for what you check?

ArsenPidhoretskyi (changes requested):

main.py
ArtMot99: If the character in the json file has a guild equal to null, then at the stage of creating an entry in the table I will get an error, here I filter out null values.

main.py
ArtMot99: Снимок экрана 2023-02-07 в 11 28 16

ArsenPidhoretskyi (approved):
Well done!

PR #244, opened 2023-02-28
Dev roman senyshyn py game models django

AndereLion:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
manjustice, line 26: - When you use `auto_now_add` - it records the time and date of creation.
- When you use `auto_now` - updates the value of the field to the current time and date every time when the object changes.
Checklist)

manjustice (changes requested):

db/models.py
AndereLion: Fixed

main.py
```diff
@@ -1,10 +1,74 @@
...
+
+        else:
+            race = Race(
+                name=players_data[key_name].get("race").get("name"),
+                description=players_data[key_name].get(
+                    "race"
+                ).get("description")
+            )
+
+        race.save()
```
YuriiKindrat, line 28: u can avoid using else condition here

main.py
```diff
@@ -1,10 +1,74 @@
...
+                    name=skill.get("name")
+                )[0]
+
+            else:
+                skill_model = Skill(
+                    name=skill.get("name"),
+                    bonus=skill.get("bonus"),
+                    race_id=race.id
+                )
+            skill_model.save()
```
YuriiKindrat, line 42: same here

main.py
```diff
@@ -1,10 +1,74 @@
...
+                )[0]
+
+            else:
+                guild = Guild(
+                    name=players_data[key_name].get("guild").get("name"),
+                    description=players_data[key_name].get(
+                        "guild"
+                    ).get("description")
+                )
+            guild.save()
```
YuriiKindrat, line 59: same

main.py
```diff
@@ -1,10 +1,74 @@
...
+        else:
+            guild = None
```
YuriiKindrat, line 61: u can create this variable before if condition

YuriiKindrat (changes requested):

main.py
AndereLion: Done

main.py
AndereLion: Fixed all.

YuriiKindrat (approved):

PR #245, opened 2023-02-28
Solution

SlavikSherbak:


main.py
```diff
@@ -1,10 +1,57 @@
...
+            Skill.objects.create(
+                name=skills.get("name"),
+                bonus=skills.get("bonus"),
+                race=race
+            )
+
+    return race
+
+
+def create_guild(guild_info: Union[None, dict[str]]) -> Guild | None:
```
Bohdan-Salamakha, line 29: use typehint like: Optional[YOUR_VALUE_TYPE] = None

main.py
```diff
@@ -1,10 +1,57 @@
...
+    if guild_info is None:
+        return None
```
Bohdan-Salamakha, line 31: just return
it returns None by default

main.py
```diff
@@ -1,10 +1,57 @@
...
+from typing import Union
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
+def create_race(race_info: dict[str, list]) -> Race:
+    if Race.objects.filter(name=race_info.get("name")).exists():
+        race = Race.objects.get(name=race_info.get("name"))
```
Bohdan-Salamakha, line 11: use get_object_or_404() to avoid Error

Bohdan-Salamakha (changes requested):
v1

main.py
SlavikSherbak: I don't understand what you mean

Grayder0152: read more [here](https://docs.djangoproject.com/en/4.1/topics/http/shortcuts/#get-object-or-404)

Grayder0152 (changes requested):

main.py
SlavikSherbak: image
I rewrote my code, but the readme advises the option that I used before

vasylhnatiuk (approved):
Well done!

PR #249, opened 2023-03-02
Solution

WDemidenko:


ZabFTFT (approved):
I was so stupid to not getting values of race, guild, skills right on the top.I like your solution =)

YuriiKindrat (approved):

db/models.py
```diff
@@ -1 +1,32 @@
...
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(
+        "describes what kind of bonus players can get from it.",
+        max_length=255
+    )
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
Serhii-Leonenko, line 15: I don't think that skill can exist without the rase

main.py
```diff
@@ -1,10 +1,46 @@
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
+import json
```
Serhii-Leonenko, line 4: Fix import according to PEP8 

https://peps.python.org/pep-0008/#imports

main.py
```diff
@@ -1,10 +1,46 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json") as players:
```
Serhii-Leonenko, line 8: what about the mode for open()?

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #250, opened 2023-03-03
Solution

AlimFreight:


main.py
```diff
@@ -1,10 +1,61 @@
...
+                    name=race_data["name"]
+            ).exists():
+                race = Race.objects.get(
+                    name=race_data["name"]
+                )
+            else:
+                race = Race.objects.create(
+                    name=race_data["name"],
+                    description=race_data["description"]
+                )
```
ArsenPidhoretskyi, line 25: You can use https://docs.djangoproject.com/en/4.1/ref/models/querysets/#get-or-create
It helps to reduce number of queries

ArsenPidhoretskyi: The same in other places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #252, opened 2023-03-03
solution

kleotan901:


main.py
```diff
@@ -1,10 +1,52 @@
 import init_django_orm  # noqa: F401
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
```
Bohdan-Salamakha, line 7: separate your function into a few functions with own logic

main.py
```diff
@@ -1,10 +1,52 @@
...
+                                  email=data[player_name]["email"],
+                                  bio=data[player_name]["bio"],
+                                  race_id=race_id,
+                                  guild_id=guild_id,
+                                  )
+        for skill in data[player_name]["race"]["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
```
Bohdan-Salamakha, line 47: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #261, opened 2023-03-07
Solution

viktoria-rybenchuk:


usrofgh (approved):

main.py
```diff
@@ -1,10 +1,76 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
+with open("players.json", "r") as sourse:
+    data = json.load(sourse)
+
 
 def main() -> None:
-    pass
+    global race_id
```
YuriiKindrat, line 12: why do u need this ?

main.py
```diff
@@ -1,10 +1,76 @@
...
+    for nickname, info in data.items():
+        if info["race"]:
+            name_race = (
+                info["race"]["name"] if
+                info["race"]["name"] else None
+            )
+            race_description = (
+                info["race"]["description"] if
+                info["race"]["description"] else None
+            )
```
YuriiKindrat, line 22: no need to make this check here, player always will have a race

main.py
```diff
@@ -1,10 +1,76 @@
...
+            if not Guild.objects.filter(name=guild_name).exists():
+                guild_object = Guild(
+                    name=guild_name,
+                    description=guild_description
+                )
+                guild_object.save()
+
+            guild_id = Guild.objects.get(name=guild_name).id
+        else:
+            guild_id = None
```
YuriiKindrat, line 51: the same as for race_ids

main.py
```diff
@@ -1,10 +1,76 @@
...
+
+        if info["race"]["skills"]:
+            for skill in info["race"]["skills"]:
+                name_skill = skill["name"] if skill["name"] else None
+                bonus = skill["bonus"] if skill["bonus"] else None
```
YuriiKindrat, line 56: it seems to me skill will always have some name and bonus )

main.py
```diff
@@ -1,10 +1,76 @@
...
+                if not Skill.objects.filter(name=name_skill).exists():
+                    skill_object = Skill(
+                        name=name_skill,
+                        bonus=bonus,
+                        race_id=race_id)
+                    skill_object.save()
+
+        email = info["email"] if info["email"] else None
+        biography = info["bio"] if info["bio"] else None
```
YuriiKindrat, line 65: email and bio also can't be blank and null accroding to the models

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):

main.py
```diff
@@ -1,10 +1,51 @@
...
-    pass
+    for nickname, info in data.items():
+        name_race = info["race"]["name"]
+        race_description = info["race"]["description"]
+
+        if not Race.objects.filter(name=name_race).exists():
+            race = Race.objects.create(
+                name=name_race,
+                description=race_description
+            )
```
Serhii-Leonenko, line 20: if race already exists, how do you get it?
For this, you can use the `get_or_create()` method. The same in other similar cases

wQuelS (changes requested):

YuriiKindrat (approved):

PR #262, opened 2023-03-07
Solution

yhwh6:


main.py
```diff
@@ -1,10 +1,51 @@
...
+    for player, info in player_info.items():
+        race_name = info["race"]["name"]
+        skills = info["race"]["skills"]
+        guild = info["guild"]["name"] if info["guild"] else None
+
+        if not Race.objects.filter(name=race_name).exists():
+            race_description = info["race"]["description"]
+            Race.objects.create(name=race_name, description=race_description)
+
+        race = Race.objects.get(name=race_name)
```
ArsenPidhoretskyi, line 20: You can use get_or_create

ArsenPidhoretskyi: The same in other places

ArsenPidhoretskyi (approved):
Good job! Also check comments for future

PR #263, opened 2023-03-07
solution

usrofgh:


DavidGlusch (approved):

db/models.py
```diff
@@ -1 +1,27 @@
...
+    name = models.CharField(unique=True, max_length=255)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+
```
YuriiKindrat, line 24: why do u need this blank line ?

main.py
```diff
@@ -1,10 +1,48 @@
...
+
+        race_name = attributes["race"]["name"]
+        if not Race.objects.filter(name=race_name).exists():
+            Race(
+                name=race_name,
+                description=attributes["race"]["description"]
+            ).save()
+
+        for skills in attributes["race"]["skills"]:
+            print(skills)
```
YuriiKindrat, line 31: remove unnesessary print here

YuriiKindrat (changes requested):

main.py
```diff
@@ -1,10 +1,47 @@
...
+                    name=skills["name"],
+                    bonus=skills["bonus"],
+                    race=Race.objects.get(name=race_name)
+                ).save()
+
+        Player(
+            nickname=name,
+            email=attributes["email"],
+            bio=attributes["bio"],
+            race=Race.objects.get(name=race_name),
```
YuriiKindrat, line 42: why do u need to make `Race.objects.get(name=race_name)` twice ? u can create variable for this

YuriiKindrat (changes requested):

main.py
```diff
@@ -1,10 +1,48 @@
...
+                    race=race
+                ).save()
+
+        Player(
+            nickname=name,
+            email=attributes["email"],
+            bio=attributes["bio"],
+            race=race,
+            guild=guild
+        ).save()
```
Serhii-Leonenko, line 45: You can use .object.create() instead of .save(). And not only in this case, try to use it everywhere

Serhii-Leonenko (changes requested):

main.py
usrofgh: I thought it's more readability, but okey, will use object.create )

main.py
```diff
@@ -1,10 +1,48 @@
...
+    for name, attributes in players_dict.items():
+        guild = attributes.get("guild")
+
+        if guild:
+            if not Guild.objects.filter(name=guild["name"]).exists():
+                Guild(
+                    name=guild["name"],
+                    description=guild["description"]
+                ).save()
+            guild = Guild.objects.get(name=guild["name"])
```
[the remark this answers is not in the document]
usrofgh: thanks, it's really more useful, didn't know about this method

.gitignore
```diff
@@ -8,3 +8,4 @@ venv/
 **__pycache__/
 *.pyc
 **db.sqlite3
+db.sqlite3
```
wQuelS, line 11: new line at the end of the file. Yeap, even here

wQuelS (changes requested):

wQuelS (approved):

PR #264, opened 2023-03-08
very dirty solution

DavidGlusch:


main.py
```diff
@@ -1,10 +1,71 @@
...
+        race_description = players_data[player]["race"]["description"]
+
+        if Race.objects.filter(name=race_name).exists():
+            Race.objects.get(name=race_name)
+
+        else:
+            Race.objects.create(
+                name=race_name,
+                description=race_description
+            )
```
kuznetsb, line 40: you can make race variable to store instance of race here like in player_guild

main.py
```diff
@@ -1,10 +1,71 @@
...
+                        )
+                    )
+
+        # Player
+
+        Player.objects.create(
+            nickname=player,
+            email=players_data[player]["email"],
+            bio=players_data[player]["bio"],
+            race=Race.objects.get(name=race_name),
```
kuznetsb, line 66: and then use race variable here

kuznetsb (approved):
Some changes I'd better do. Else seems okay

main.py
```diff
@@ -1,10 +1,71 @@
...
+
+            if Guild.objects.filter(name=guild_name).exists():
+                player_guild = Guild.objects.get(name=guild_name)
+            else:
+                player_guild = Guild.objects.create(
+                    name=guild_name,
+                    description=guild_description
+                )
+        else:
+            player_guild = None
```
YuriiKindrat, line 27: no need to use else here, u can create this variable before if condition

main.py
```diff
@@ -1,10 +1,71 @@
...
+
+        # Race
+        race_name = players_data[player]["race"]["name"]
+        race_description = players_data[player]["race"]["description"]
+
+        if Race.objects.filter(name=race_name).exists():
+            Race.objects.get(name=race_name)
```
YuriiKindrat, line 34: why do u need to do this ?

main.py
YuriiKindrat, line 66: u are using Race.objects.get(name=race_name) for a few times in ur code, it is better to create a variable for this

YuriiKindrat (changes requested):
also try to use get_or_create method, it will help to optimize ur solution

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
YuriiKindrat, line 26: image
didn't see this before

main.py
```diff
@@ -1,10 +1,61 @@
...
+            guild_name = players_data[player]["guild"]["name"]
+            guild_description = players_data[player]["guild"]["description"]
+
+            if Guild.objects.filter(name=guild_name).exists():
+                player_guild = Guild.objects.get(name=guild_name)
+            else:
+                player_guild = Guild.objects.create(
+                    name=guild_name,
+                    description=guild_description
+                )
```
YuriiKindrat, line 26: no need to use else here, swap ur conditions

YuriiKindrat (changes requested):
also remove all unnecessary comments

YuriiKindrat (approved):

PR #268, opened 2023-03-09
Solution

NEkropo1:


oksana-feshchenko (approved):

main.py
```diff
@@ -1,10 +1,73 @@
 import init_django_orm  # noqa: F401
+import json
 
+from django.db import transaction
 from db.models import Race, Skill, Player, Guild
```
YuriiKindrat, line 5: image

main.py
```diff
@@ -1,10 +1,73 @@
...
+
+
+def create_guild(guild_data: dict) -> Guild:
+    if guild_data:
+        guild, created = Guild.objects.get_or_create(
+            name=guild_data["name"],
+            defaults={"description": guild_data.get("description")},
+        )
+    else:
+        guild = None
```
YuriiKindrat, line 29: u can avoid using else here

main.py
```diff
@@ -1,10 +1,73 @@
...
+    player, created = Player.objects.get_or_create(
+        nickname=player_data["nickname"],
+        defaults={
+            "email": player_data["email"],
+            "bio": player_data["bio"],
+            "race": race,
+            "guild": guild,
+        },
+    )
+    return player
```
YuriiKindrat, line 61: why do u need return here, if u don't use it later ?

YuriiKindrat (changes requested):

main.py
NEkropo1: Yup that was made for future purpose either we want to manipulate created player further, no need here

main.py
NEkropo1: Don't actually think you should do this within 2 fields, but for general practice, agree.

YuriiKindrat (approved):

PR #273, opened 2023-03-09
solution

Yulia-Mala:


main.py
```diff
@@ -1,10 +1,38 @@
...
+                    race=Race.objects.get(name=info["race"]["name"]))
+        player_guild = (None if info["guild"] is None
+                        else Guild.objects.get(name=info["guild"]["name"]))
+        Player.objects.create(
+            nickname=name,
+            email=info["email"],
+            bio=info["bio"],
+            race=Race.objects.get(name=info["race"]["name"]),
+            guild=player_guild
+        )
```
Litvinenko23, line 35: I'm not sure, but I think mentor will reject this decision. Yaroslav said that a function should not contain too many lines of code. As an example, he said 8 rows, but I think it's not correct in this case.

Litvinenko23 (changes requested):

Litvinenko23 (approved):
Sasha said that it is possible in one function

main.py
```diff
@@ -1,10 +1,38 @@
...
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=Race.objects.get(name=info["race"]["name"]))
+        player_guild = (None if info["guild"] is None
+                        else Guild.objects.get(name=info["guild"]["name"]))
+        Player.objects.create(
+            nickname=name,
+            email=info["email"],
+            bio=info["bio"],
+            race=Race.objects.get(name=info["race"]["name"]),
```
YuriiKindrat, line 33: u are using Race.objects.get(name=info["race"]["name"]) twice here why don't u create variable for this ?

main.py
```diff
@@ -1,10 +1,39 @@
...
+    with open("players.json", "r") as file:
+        players = json.load(file)
+    for name, info in players.items():
+        if not Race.objects.filter(name=info["race"]["name"]).exists():
+            Race.objects.create(
+                name=info["race"]["name"],
+                description=info["race"]["description"])
+        player_race = Race.objects.get(name=info["race"]["name"])
+        if (info["guild"] is not None and not
+                Guild.objects.filter(name=info["guild"]["name"]).exists()):
```
YuriiKindrat, line 18: fix indentation here

main.py
```diff
@@ -1,10 +1,39 @@
...
+            Guild.objects.create(
+                name=info["guild"]["name"],
+                description=info["guild"]["description"])
```
YuriiKindrat, line 21: same

main.py
```diff
@@ -1,10 +1,39 @@
...
+        for skill in info["race"]["skills"]:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=player_race)
```
YuriiKindrat, line 27: same

YuriiKindrat (changes requested):

main.py
Yulia-Mala: with any other indentation except this one PyCharm show me PEP8 warning. what exactly indentation do you recommend  ?

main.py
Yulia-Mala: if you meant parentheses - done

main.py
Yulia-Mala: if you meant parentheses - done

main.py
```diff
@@ -1,10 +1,41 @@
...
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players = json.load(file)
+    for name, info in players.items():
+        if not Race.objects.filter(name=info["race"]["name"]).exists():
+            Race.objects.create(
+                name=info["race"]["name"],
+                description=info["race"]["description"])
+        player_race = Race.objects.get(name=info["race"]["name"])
```
Serhii-Leonenko, line 16: This will not affect the number of database queries but may improve readability. For this, use the `get_or_create()` method. Just try it. The same in other similar cases

Serhii-Leonenko (changes requested):

main.py
Yulia-Mala: Thanks. It's better now

main.py
```diff
@@ -1,10 +1,40 @@
...
+            name=info["race"]["name"],
+            description=info["race"]["description"]
+        )
+        if info["guild"] is not None:
+            player_guild, _ = Guild.objects.get_or_create(
+                name=info["guild"]["name"],
+                description=info["guild"]["description"]
+            )
+        else:
+            player_guild = None
```
YuriiKindrat, line 23: no need to use else here, u can create this variable at the start of loop

YuriiKindrat (approved):

PR #276, opened 2023-03-09
Solution

aarrtemm:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
```
vasylhnatiuk, line 5: name must be a unique char field, here and for all of these models.

main.py
```diff
@@ -1,10 +1,49 @@
...
 def main() -> None:
-    pass
+    with open("players.json", "r") as date:
+        player_info = json.load(date)
+    for player, info in player_info.items():
+        race_name = info["race"]["name"]
+        skills = info["race"]["skills"]
+        guild = info["guild"] if info["guild"] else None
+
+        if not Race.objects.filter(name=race_name).exists():
```
vasylhnatiuk, line 14: You can use `get_or_create` if objects don't exit https://docs.djangoproject.com/en/4.1/ref/models/querysets/#get-or-create
here and other places

vasylhnatiuk (changes requested):
In general, it would be better to separate the main function into a few functions with different responsibilities

ArsenPidhoretskyi (approved):
Well done!

PR #283, opened 2023-03-09
Solution

oksanaaam:


main.py
```diff
@@ -1,10 +1,52 @@
...
+        guild = player_info.get("guild")
+
+        if Race.objects.filter(name=race["name"]).exists():
+            race = Race.objects.get(name=race["name"])
+        else:
+            race = Race.objects.create(name=race["name"],
+                                       description=race["description"])
+
+        for skill in skills:
+            if not Skill.objects.filter(name=skill["name"]).exists():
```
NEkropo1, line 24: You could get a key error, or value error, depending on which field is not provided in `skills = player_info.get("race")["skills"]` 
I mean, if you already using get() for checking if we have race field in code you could refactor a little bit this code, providing same check for skills field...

main.py
```diff
@@ -1,10 +1,52 @@
...
+                skill = Skill.objects.create(
+                    name=skill.get("name"),
+                    bonus=skill.get("bonus"),
+                    race=race)
+            else:
+                Skill.objects.get(name=skill["name"])
```
NEkropo1, line 30: Bonus for these fields, you can simply use get_or_create instead whole check, but it's just a suggestion:
[https://www.queworx.com/django/django-get_or_create/](url)

NEkropo1 (changes requested):

main.py
```diff
@@ -1,10 +1,46 @@
...
+                race=race
+            )
+
+        if guild:
+            guild, created = Guild.objects.get_or_create(
+                name=guild["name"],
+                description=guild["description"]
+            )
+        else:
+            guild = None
```
YuriiKindrat, line 35: no need to use else here, u can use get method for guild variable before

YuriiKindrat (changes requested):

PR #289, opened 2023-03-09
Solution

NazarHladaniuk:


myroslavatrschk (approved):

main.py
```diff
@@ -1,10 +1,41 @@
...
+                    race=race_get
+                )
+
+        if player_info.get("guild"):
+            guild_get = Guild.objects.get_or_create(
+                name=player_info["guild"]["name"],
+                description=player_info["guild"]["description"]
+            )[0]
+        else:
+            guild_get = None
```
YuriiKindrat, line 30: no need to use else here, u can create this variable before if condition

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #290, opened 2023-03-09
'Solution'

alexlega:


main.py
```diff
@@ -1,10 +1,59 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+
+    for user, info in data.items():
+        nickname = user
+        email = info["email"] if info["email"] else None
```
ZhAlexR, line 12: You can use "info.get(email)" instead and if no email the `None` will be returned by default

main.py
```diff
@@ -1,10 +1,59 @@
...
+        bio = info["bio"] if info["bio"] else None
+
+        if info["guild"]:
+            guild_name = (info["guild"]["name"]
```
ZhAlexR, line 16: You have already checked that guild has some info. Why do you need to check that one more time with "name"?

ZhAlexR (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
alexlega: yeah, you are right, my mistake, thanks!

main.py
alexlega: I thought that we might have some information about the "guild" and then we check the items inside the "guild"

alexlega: But maybe I'm wrong

main.py
```diff
@@ -1,10 +1,59 @@
 import init_django_orm  # noqa: F401
-
+import json
 from db.models import Race, Skill, Player, Guild
```
Serhii-Leonenko, line 3: Fix imports according to PEP8

https://peps.python.org/pep-0008/#imports

db/models.py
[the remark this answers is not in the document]
Serhii-Leonenko: Player shouldn't be deleted if the guild is deleted, but it should be deleted if we delete the race

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
Serhii-Leonenko, line 24: race can't be None

main.py
```diff
@@ -1,10 +1,59 @@
...
+            guild_name = (info["guild"]["name"]
+                          if info["guild"]["name"] else None)
+            guild_description = (info["guild"]["description"]
+                                 if info["guild"]["description"] else None)
+
+            if not Guild.objects.filter(name=guild_name).exists():
+                guild_link = Guild.objects.create(
+                    name=guild_name,
+                    description=guild_description
+                )
```
Serhii-Leonenko, line 25: This will not affect the number of database queries but may improve readability. For this, use the get_or_create() method. Just try it )

main.py
```diff
@@ -1,10 +1,59 @@
...
+    for user, info in data.items():
+        nickname = user
+        email = info.get("email")
+        bio = info.get("bio")
+
+        if info["guild"]:
+            guild_name = (info["guild"]["name"]
+                          if info["guild"]["name"] else None)
+            guild_description = (info["guild"]["description"]
+                                 if info["guild"]["description"] else None)
```
Serhii-Leonenko, line 19: You can use dict.get() method instead of ternary and try to remove redundant variables

Serhii-Leonenko (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL)
```
YuriiKindrat, line 24: player can't exist without race, fix this
image

main.py
```diff
@@ -1,10 +1,53 @@
...
+                        bonus=skill_bonus,
+                        race=race_link
+                    )
+
+        Player.objects.create(
+            nickname=nickname,
+            email=email,
+            bio=bio,
+            race=race_link if info["race"]["name"] else None,
+            guild=guild_link if info.get("guild") else None
```
YuriiKindrat, line 49: no need to use if/else condition for get method

main.py
```diff
@@ -1,10 +1,53 @@
...
+                        name=skill_name,
+                        bonus=skill_bonus,
+                        race=race_link
+                    )
+
+        Player.objects.create(
+            nickname=nickname,
+            email=email,
+            bio=bio,
+            race=race_link if info["race"]["name"] else None,
```
YuriiKindrat, line 48: no need to make check here, player always will have a race

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #291, opened 2023-03-09
Solution

VladyslavCherkashyn:


inrock1 (approved):
good

db/models.py
```diff
@@ -1 +1,30 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255, help_text="skill bonus type")
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
Serhii-Leonenko, line 12: I think it's better to use on_delete=models.CASCADE, without null=True, because we don't need skills that exist without race

main.py
```diff
@@ -1,10 +1,44 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
```
Serhii-Leonenko, line 5: Do you need a blank line between these imports?

main.py
```diff
@@ -1,10 +1,44 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
```
Serhii-Leonenko, line 9: What about the mode for open?)

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #295, opened 2023-03-09
Develop

ZhAlexR:


db/models.py
```diff
@@ -1 +1,41 @@
...
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "description provided by a user about himself/herself",
+        max_length=255
+    )
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.SET_NULL,
+        null=True
+    )
```
YuriiKindrat, line 35: image
player should always have race

db/models.py
```diff
@@ -1 +1,41 @@
...
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.SET_NULL,
+        null=True
+    )
```
YuriiKindrat, line 16: image
are u sure that skill can have no race ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #296, opened 2023-03-09
Solution

Zenionoi:


db/models.py
```diff
@@ -1 +1,33 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race,
+                             on_delete=models.CASCADE,
+                             related_name="skill_set")
```
wQuelS, line 14: why did you provide a default value for `related_name` arg ?

wQuelS (changes requested):

db/models.py
Zenionoi: Firstly, I gave it a `skill` value, but one test not passed and i changed `related_name`,not deleted

YuriiKindrat (approved):

PR #311, opened 2023-03-13
Solution

nazar-dmytrenko:


main.py
```diff
@@ -1,10 +1,44 @@
...
+        if guild is not None:
+            guild = (Guild.objects.get_or_create(
+                name=guild["name"],
+                description=guild["description"]
+            ))[0]
+
+        Player.objects.get_or_create(
+            nickname=nickname,
+            email=data["email"],
+            bio=data["bio"],
```
Bohdan-Salamakha, line 38: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #312, opened 2023-03-14
Solution

StreetNik:


main.py
```diff
@@ -1,10 +1,61 @@
...
+                name=race_data["name"]
+        ).exists() is False:
+            race = Race.objects.create(
+                name=race_data["name"],
+                description=race_data["description"]
+            )
+        else:
+            race = Race.objects.get(
+                name=race_data["name"]
+            )
```
ArsenPidhoretskyi, line 24: Use get_or_create
The same with other places

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,40 @@
...
+
+        Player.objects.create(
+            nickname=player,
+            email=dict_players[player]["email"],
+            bio=dict_players[player]["bio"],
+            race=Race.objects.get_or_create(
+                name=race_data["name"],
+                description=race_data["description"]
+            )[0],
+            guild=None if guild_data is None else Guild.objects.get_or_create(
```
Grayder0152, line 24:
```suggestion
            guild=guild_data if guild_data is None else Guild.objects.get_or_create(
```

Grayder0152 (approved):

PR #317, opened 2023-03-21
Solution

oleksandr-lysanets:


main.py
```diff
@@ -1,10 +1,53 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
+with open("players.json") as players:
+    players_data = json.load(players)
```
Bohdan-Salamakha, line 9: place it into a separated function

main.py
```diff
@@ -1,10 +1,53 @@
...
+            else:
+                guild = Guild.objects.get(name=data["guild"]["name"])
+
+        Player.objects.create(
+            nickname=name,
+            email=data["email"],
+            bio=data["bio"],
+            race=race,
+            guild=guild
+        )
```
Bohdan-Salamakha, line 50: separate your function into a few functions with own logic

Bohdan-Salamakha (changes requested):
v1

main.py
```diff
@@ -1,10 +1,69 @@
...
+
+def create_skill(data: dict) -> None:
+    for skill in data["race"]["skills"]:
+        if not Skill.objects.filter(name=skill["name"]).exists():
+            skill = Skill.objects.create(
+                name=skill["name"],
+                bonus=skill["bonus"]
+            )
+        else:
+            skill = Skill.objects.get(name=skill["name"])
```
olena-hul, line 32: Use `.get_or_create` method for this

olena-hul: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#get-or-create

olena-hul: Fix in all places where needed

olena-hul (changes requested):

main.py
```diff
@@ -1,10 +1,60 @@
...
+    return race
+
+
+def create_skill(data: dict) -> None:
+    for skill in data["race"]["skills"]:
+        skill, _ = Skill.objects.get_or_create(
+            name=skill["name"],
+            bonus=skill["bonus"]
+        )
+        skill.race = create_race(data)
```
olena-hul, line 27: Create race once at the begginning of the function
And then just assign the value in a loop

This way you'll reduce some queries to the database

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #318, opened 2023-03-23
Solution

lotrik84:


main.py
```diff
@@ -1,10 +1,58 @@
...
+                description=players[player]["guild"]["description"]
+            )
+        for skill in players[player]["race"]["skills"]:
+            if not Skill.objects.filter(name=skill["name"]):
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=Race.objects.get(
+                        name=players[player]["race"]["name"]
+                    )
```
ArsenPidhoretskyi, line 42: The same query in loop so bad...

main.py
```diff
@@ -1,10 +1,58 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        players = json.load(file)
+
+    for player in players:
```
ArsenPidhoretskyi, line 12: You can use .items

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,57 @@
...
+                    bonus=skill["bonus"],
+                    race=players_race
+                )
+        if not is_player_exist:
+            Player.objects.create(
+                nickname=player,
+                email=info["email"],
+                bio=info["bio"],
+                race=players_race,
+                guild=Guild.objects.get(
```
vasylhnatiuk, line 51: it not a good idea to get a guild in create(), it is hard to read. write it abowe

vasylhnatiuk (approved):
Good job!

PR #319, opened 2023-03-24
Develop

ZaikaDaria:


main.py
```diff
@@ -1,10 +1,43 @@
...
+        data = json.load(file)
+
+        for player_name, player_info in data.items():
+            race = player_info.get("race")
+            guild = player_info.get("guild")
+            skills = race.get("skills")
+
+            race, _ = Race.objects.get_or_create(
+                name=race["name"],
+                description=race["description"]
```
ArsenPidhoretskyi, line 18: Race can be None due to .get method use player_info["race"] or .get but with default argument as dict
The same with skill and player

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,44 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        data = json.load(file)
+
+        for player_name, player_info in data.items():
```
katevasilenko, line 11: Do not overload context manager

db/models.py
```diff
@@ -1 +1,31 @@
...
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True
+    )
+    created_at = models.DateTimeField(auto_now=True)
```
katevasilenko, line 31: I believe you need `auto_now_add` here

main.py
```diff
@@ -1,10 +1,44 @@
...
+                    description=guild["description"]
+                )
+            if skills:
+                for skill in skills:
+                    Skill.objects.get_or_create(
+                        name=skill["name"],
+                        bonus=skill["bonus"],
+                        race=race
+                    )
+            if not None:
```
katevasilenko, line 34: I don’t get this? What do you try to check here

katevasilenko (changes requested):

main.py
ZaikaDaria: "Race can be None due to .get method use player_info["race"] or .get but with default argument as dict
The same with skill and player"
Worked with Arsen's suggestions.

main.py
katevasilenko: I meant this https://github.com/mate-academy/py-game-models/blob/master/checklist.md#1-do-not-overload-the-context-manager

katevasilenko (changes requested):

katevasilenko (approved):
Looks good to me now

PR #323, opened 2023-04-01
solution

MkrtychKhachatrian:


main.py
```diff
@@ -1,10 +1,68 @@
...
-    pass
+    with open("players.json", "r") as file:
+        players_data = json.load(file)
+
+    for player_nickname, player in players_data.items():
+
+        # Creating a race
+
+        race_name = player.get("race").get("name")
+        race_description = player.get("race").get("description")
```
ArsenPidhoretskyi, line 18: First get is not needed use brackets or do as follow:
player.get("race", dict()).get("name")
Fin in other places

main.py
```diff
@@ -1,10 +1,68 @@
...
+        if not Race.objects.filter(
+                name=race_name).exists() and race_description:
+            race_obj = Race.objects.create(
+                name=race_name,
+                description=race_description
+            )
+        elif not Race.objects.filter(
+                name=race_name).exists() and race_description is None:
+            race_obj = Race.objects.create(name=race_name)
```
ArsenPidhoretskyi, line 27: 3 queries always as minimum. You can use get_or_create
Fix in other places

ArsenPidhoretskyi (changes requested):

main.py
khachatrian2000: Great method, thanks!

vasylhnatiuk (approved):
Good job!

PR #325, opened 2023-04-01
Solution

olenaliuby:


db/models.py
```diff
@@ -1 +1,58 @@
...
+class Skill(models.Model):
+    name = models.CharField(unique=True, max_length=255)
+    bonus = models.CharField(
+        "Describes what kind of bonus players can get from it",
+        max_length=255
+    )
+    race = models.ForeignKey(
+        "Race",
+        related_name="skill_set",
+        on_delete=models.DO_NOTHING
```
katevasilenko, line 21: I believe skills can’t live without race so you need to change the deleting type
Also thy to avoid `DO_NOTHING`

db/models.py
```diff
@@ -1 +1,58 @@
...
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "Stores a short description provided by a user about himself/herself.",
+        max_length=255
+    )
+    race = models.ForeignKey(
+        "Race",
+        related_name="player_race",
+        on_delete=models.DO_NOTHING
```
katevasilenko, line 46: B77A2895-A83B-4192-AE6F-A968A455F735
Choose another deleting type to match the task

db/models.py
```diff
@@ -1 +1,58 @@
...
+    )
+    guild = models.ForeignKey(
+        "Guild",
+        related_name="player_guild",
+        on_delete=models.SET_NULL,
+        null=True
+    )
+    created_at = models.DateTimeField(auto_now=True)
```
katevasilenko, line 54: You need `auto_now_add`

db/models.py
```diff
@@ -1 +1,58 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.nickname}, {self.email}. {self.bio}. " \
+               f"Race: {self.race}. Guild: {self.guild}."
```
katevasilenko, line 58: Please use brackets to avoid \ symbol

main.py
```diff
@@ -1,11 +1,44 @@
...
+            race=race,
+            guild=guild
+        ) if not Player.objects.filter(
+            nickname=nickname
+        ).exists() else None
 
 
 if __name__ == "__main__":
     main()
+    print(Player.objects.all())
```
katevasilenko, line 44: Please remove this from the final version

katevasilenko (changes requested):

db/models.py
olenaliuby: Can we use PROTECTED in this case? Because, if we delete the race — users can be deleted too. So, It's better if we have ability to delete the race only if it doesn't have any referenced users.

main.py
olenaliuby: Thank you for helping and highlighting all the issues 👍🏻

vasylhnatiuk (approved):
Well done!

PR #329, opened 2023-04-06
Solution

Bloodviel:


main.py
```diff
@@ -1,10 +1,42 @@
...
+
+        skills = info["race"]["skills"]
+        for skill in skills:
+            Skill.objects.get_or_create(
+                name=skill["name"],
+                bonus=skill["bonus"],
+                race=race_obj
+            )
+
+        guild = info["guild"]
```
ArsenPidhoretskyi, line 26: As a result inside player must be a guild or None, not dict
If we have guild equal to dict it will be so bad

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #332, opened 2023-04-13
Solution

Roman28101:


main.py
```diff
@@ -1,10 +1,42 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        data = json.load(file)
+
+    for key, value in data.items():
+        race_data = value["race"] if value["race"] else None
+        if not Race.objects.filter(name=race_data["name"]).exists():
+            Race.objects.create(
```
ArsenPidhoretskyi, line 15: Use get_or_create

main.py
```diff
@@ -1,10 +1,42 @@
...
+                name=race_data["name"],
+                description=race_data["description"])
+            for skill in race_data["skills"]:
+                print(skill)
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=Race.objects.get(name=race_data["name"]))
```
ArsenPidhoretskyi, line 23: Unoptimized code... In each loop go to db and get race is so bad

main.py
```diff
@@ -1,10 +1,42 @@
...
+
+        guild_data = value["guild"] if value["guild"] else None
+        if value["guild"] is not None:
+            if not Guild.objects.filter(name=guild_data["name"]).exists():
+                Guild.objects.create(
```
ArsenPidhoretskyi, line 28: Use get_or_create

main.py
```diff
@@ -1,10 +1,42 @@
...
+                    name=guild_data["name"],
+                    description=guild_data["description"])
+
+        if not Player.objects.filter(nickname=key).exists():
+            Player.objects.create(
+                nickname=key,
+                email=value["email"],
+                bio=value["bio"],
+                race=Race.objects.get(name=race_data["name"]),
+                guild=Guild.objects.get(name=guild_data["name"])
```
ArsenPidhoretskyi, line 38: You can already have race and guild. Please no need so much times go inside db

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -11,32 +11,27 @@ def main() -> None:
...
+                bonus=skill["bonus"],
+                race=race_name)
+
+        guild_data = value["guild"]
+        Player.objects.get_or_create(
+            nickname=key,
+            email=value["email"],
+            bio=value["bio"],
+            race=race_name,
+            guild=Guild.objects.get_or_create(
```
vasylhnatiuk, line 31: it's not a good idea to get the guild in Player.objects.get_or_create, move this code above instead of guild_data = value["guild"]

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Well done

PR #334, opened 2023-04-14
v1

NVMBR13:


main.py
```diff
@@ -1,10 +1,38 @@
...
+        players = json.load(file)
+
+    for name, info in players.items():
+        guild = info.get("guild")
+        if guild:
+            guild, _ = Guild.objects.get_or_create(
+                name=guild.get("name"),
+                description=guild.get("description"))
+
+        race = info.get("race")
```
ArsenPidhoretskyi, line 18: Use brackets:
race = info["race"]
 or raise error is race is None

ArsenPidhoretskyi: The same with skills

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #338, opened 2023-04-24
Solution +

MaksNochvai:


main.py
```diff
@@ -1,10 +1,41 @@
...
+        )
+
+        for skill in race["skills"]:
+            Skill.objects.get_or_create(
+                race=new_race,
+                name=skill["name"],
+                bonus=skill["bonus"]
+            )
+
+        guild = info.get("guild", None)
```
olena-hul, line 26: No need to pass None. It's the default value that is used when key is not present in the dictionary

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #343, opened 2023-05-04
Solution

nickkozlov90:


EdAlekseiev (changes requested):
Fix tests, please

main.py
```diff
@@ -1,10 +1,57 @@
...
+        if player_guild:
+            if not Guild.objects.filter(name=player_guild["name"]).exists():
+                if player_guild["description"]:
+                    guild_description = player_guild["description"]
+                else:
+                    guild_description = None
+                Guild.objects.create(
+                    name=player_guild["name"],
+                    description=guild_description
+                )
```
olena-hul, line 25: you can use get_or_create method https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create

olena-hul (changes requested):

main.py
```diff
@@ -1,10 +1,53 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        players = json.load(file)
+
+    for player_name, player_data in players.items():
+        player_guild = player_data.get("guild")
+        player_race_name = player_data.get("race").get("name")
```
ArsenPidhoretskyi, line 13: It can return Exception None Type. Use:
```suggestion
        player_race_name = player_data.get("race", {}).get("name")
```
Fix in other places

main.py
```diff
@@ -1,10 +1,53 @@
...
+
+        if player_guild is not None:
+            guild = Guild.objects.get_or_create(
+                name=player_guild.get("name"),
+                defaults={
+                    "description": player_guild.get("description")
+                }
+            )[0]
+        else:
+            guild = None
```
ArsenPidhoretskyi, line 25:
```suggestion
        guild = None
        if player_guild is not None:
            guild, _ = Guild.objects.get_or_create(
                name=player_guild.get("name"),
                defaults={
                    "description": player_guild.get("description")
                }
            )
```

ArsenPidhoretskyi: The same formating in other places

main.py
```diff
@@ -1,10 +1,53 @@
...
+            }
+        )[0]
+
+        for skill in player_skills:
+            if not Skill.objects.filter(name=skill["name"]).exists():
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race
+                )
```
ArsenPidhoretskyi, line 42: Use get_or_create also

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,52 @@
...
+                "description": player_race_description
+            }
+        )
+
+        for skill in player_skills:
+            Skill.objects.get_or_create(
+                name=skill["name"],
+                bonus=skill["bonus"],
+                defaults={"race": race}
+            )
```
ArsenPidhoretskyi, line 41: But if one race(Elf) will have for example skill with name "Berserk" with bonus = 2
And another(Dwarf) also want to have skill with name "Berserk" and bonus = 2
You will not create skill for dwarf. And it will be incorrect

ArsenPidhoretskyi (changes requested):
Well done!

main.py
nickkozlov90: In task it's said that "Each race has unique skills." That's why I consider that two races can not have skills with the same name.

ArsenPidhoretskyi (approved):
Well done!

PR #347, opened 2023-05-07
'Solution'

IvanGuculyak:


main.py
```diff
@@ -1,10 +1,52 @@
...
+
+    for player_name, information in players_dict.items():
+
+        race_name = information["race"]["name"]
+        race_description = information["race"]["description"]
+        if not Race.objects.filter(name=race_name).exists():
+            Race.objects.create(
+                name=race_name,
+                description=race_description
+            )
```
Grayder0152, line 20: Instead of this use [get_or_create](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create)

Grayder0152 (changes requested):

main.py
```diff
@@ -1,10 +1,49 @@
...
+            Guild.objects.get_or_create(
+                name=guild_name,
+                description=guild_description
+            )
+
+        Player.objects.create(
+            nickname=player_name,
+            email=information["email"],
+            bio=information["bio"],
+            race=Race.objects.get(name=race_name),
```
ArsenPidhoretskyi, line 42: Why you need to get it you can do in:
            name=race_name,
            description=race_description
        )

ArsenPidhoretskyi: The same with guild

main.py
```diff
@@ -1,10 +1,49 @@
...
+        race_name = information["race"]["name"]
+        race_description = information["race"]["description"]
+        Race.objects.get_or_create(
+            name=race_name,
+            description=race_description
+        )
+        for skill in information["race"]["skills"]:
+            skill_name = skill["name"]
+            skill_bonus = skill["bonus"]
+            skill_race = Race.objects.get(name=race_name)
```
ArsenPidhoretskyi, line 23: Why you need to get it inside loop get_or_create return instance

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,41 @@
...
+                name=information["guild"]["name"],
+                description=information["guild"]["description"]
+            )
+
+        Player.objects.create(
+            nickname=player_name,
+            email=information["email"],
+            bio=information["bio"],
+            race=race,
+            guild=guild if guild is not None else None
```
ArsenPidhoretskyi, line 37:
```suggestion
            guild=guild
```
This check is so strange if guild is None place None if no place guild

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #349, opened 2023-05-10
Develop

VladyslavOladko:


main.py
```diff
@@ -1,10 +1,45 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json") as f:
+        data = json.load(f)
+
+    for nickname, player_data in data.items():
+        race_data = player_data["race"]
+        race, created = Race.objects.get_or_create(
```
ArsenPidhoretskyi, line 14:
```suggestion
        race, _ = Race.objects.get_or_create(
```
You do not need info about created. In this case you can use _

ArsenPidhoretskyi: The same with guild

main.py
```diff
@@ -1,10 +1,45 @@
...
+            )
+
+        guild_data = player_data["guild"]
+        if guild_data:
+            guild, created = Guild.objects.get_or_create(
+                name=guild_data["name"],
+                description=guild_data["description"]
+            )
+        else:
+            guild = None
```
ArsenPidhoretskyi, line 33:
```suggestion
        guild = None
        if guild_data:
            guild, created = Guild.objects.get_or_create(
                name=guild_data["name"],
                description=guild_data["description"]
            )
```
Better to write like this)

db/models.py
```diff
@@ -1 +1,32 @@
...
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        null=True
+    )
+    created_at = models.DateTimeField(default=datetime.datetime)
```
ArsenPidhoretskyi, line 32: Read about auto_now_add=True and use it instead of default
Also check about auto_now and difference between them)

main.py
ArsenPidhoretskyi, line 33:
```suggestion
        guild = None
        if guild_data:
            guild, created = Guild.objects.get_or_create(
                name=guild_data["name"],
                description=guild_data["description"]
            )
```
Better to write like this)

ArsenPidhoretskyi (changes requested):

PR #366, opened 2023-05-29
Solution

KittNeKit:


main.py
```diff
@@ -1,10 +1,49 @@
...
-    pass
+    with open("players.json") as conf:
+        players = json.load(conf)
+    for player, info in players.items():
+        Player.objects.create(
+            nickname=player,
+            email=info["email"],
+            bio=info["bio"],
+            race=race_create(info["race"]),
+            guild=guild_create(info["guild"]),
```
olena-hul, line 45: Use .get() everywhere to avoid KeyError

main.py
```diff
@@ -1,10 +1,49 @@
...
+                race=race
+            )
+    return race
+
+
+def guild_create(guild_data: Optional[dict] = None) -> Guild | None:
+    if guild_data is None:
+        return
+    guild_name = guild_data["name"]
+    guild, created = Guild.objects.get_or_create(
```
olena-hul, line 29: The variable `created` is not used. You can use `_` instead of variable names in cases like this - then it won't be highlighted as an unused variable in the IDE

olena-hul (changes requested):

main.py
```diff
@@ -1,10 +1,49 @@
...
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
+def race_create(player_race_data: dict) -> Race:
+    race_name = player_race_data["name"]
+    race, created = Race.objects.get_or_create(
+        name=race_name,
+        description=player_race_data["description"]
```
olena-hul, line 13: .get() is not used everywhere

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #367, opened 2023-05-30
Develop

IvanKorshunovE:


ArsenPidhoretskyi (changes requested):
You do so much queries. You can return from function objects... No need to make so much queries to database

Grayder0152 (approved):

PR #372, opened 2023-06-04
solution

alim2709:


main.py
```diff
@@ -1,10 +1,51 @@
...
+import json
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        info_for_players = json.load(file)
+    for name in info_for_players:
+        Race.objects.get_or_create(
+            name=info_for_players.get(name).get("race")["name"],
```
olena-hul, line 12: Use variables instead of duplicating values retrieval

main.py
```diff
@@ -1,10 +1,50 @@
...
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        info_for_players = json.load(file)
+    for name in info_for_players:
+        race_info = info_for_players.get(name).get("race")
+        Race.objects.get_or_create(
+            name=race_info.get("name"),
+            description=race_info.get("description")
+        )
```
EdAlekseiev, line 15: get_or_create returns created object. You can save the result of get_or_create to a variable and use it when you need instead of duplicating values retrieval

main.py
```diff
@@ -1,10 +1,45 @@
...
+            email=info_for_players.get(name).get("email"),
+            bio=info_for_players.get(name).get("bio"),
+            race=race_object,
+            guild=(
+                Guild.objects.get(
+                    name=guild_info.get("name")
+                )
+                if guild_info is not None
+                else None
+            )
```
EdAlekseiev, line 41: The same issue as with Race

EdAlekseiev (changes requested):

olena-hul (approved):

PR #379, opened 2023-06-12
Solution

markus2812:


db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
...
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Guild',
```
vasylhnatiuk, line 16: change ' to ", use Black

vasylhnatiuk: here and in all of files

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #388, opened 2023-06-24
Solution

dmitriy-kds:


YukihiroSM (approved):
Good job!
In future use black formatter to fix formatting isses in code. Like `'` instead of `"` in migrations.

PR #391, opened 2023-06-28
Initial implementation

thorium-zero:


IhorVoskoboinikov (approved):
As for me your code is well) I run your code and PyCharm say to me that have to use this annotation: Optional[Skill] (import from typing) if you return None from function.

main.py
```diff
@@ -1,10 +1,55 @@
...
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
+def get_race(race_info: dict) -> Race:
+    result = Race.objects.get_or_create(
+        name=race_info["name"],
+        description=race_info["description"]
+    )[0]
```
YuriiKindrat, line 11: result is bad name for variable, also it is better to use underscore here, instead of indexes ))

YuriiKindrat: same is for similar cases

YuriiKindrat (changes requested):

main.py
```diff
@@ -1,10 +1,53 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
+def get_race(race_info: dict) -> Race:
+    race, _ = Race.objects.get_or_create(
+        name=race_info["name"],
+        description=race_info["description"]
+    )
+    for skill_info in race_info["skills"]:
+        _ = get_skill(skill_info, race)
```
Oleg-s-s, line 13: Why you create a var here?

main.py
```diff
@@ -1,10 +1,53 @@
...
+    )
+    return skill
+
+
 def main() -> None:
-    pass
+    with open("players.json", "r") as data_source:
+        players = json.load(data_source)
+    for player_name, player_data in players.items():
+        _ = Player.objects.get_or_create(
```
Oleg-s-s, line 42: And here

Oleg-s-s (approved):

main.py
thorium-zero: @Oleg-s-s Discards are placeholder variables that are intentionally unused in application code. Discards make the intent of your code clear. A discard indicates that our code never uses the variable. They enhance its readability and maintainability.

main.py
thorium-zero: same here

PR #395, opened 2023-06-29
solution

oleksiikolii:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
ant-komarov, line 26: You should use auto_now_add. If You use auto_now this field will be changed when you apply some updates to player

main.py
```diff
@@ -1,10 +1,65 @@
...
+                race_id=get_race_id(race)
+            )
+
+
+def create_race(race: dict) -> None:
+    if not Race.objects.filter(name=race["name"]).exists():
+        Race.objects.create(
+            name=race["name"],
+            description=race["description"]
+        )
```
ant-komarov, line 62: Probably, it will be more logic to define these functions before using them in "main"-function

ant-komarov (changes requested):

ant-komarov (approved):
Good job👍

PR #397, opened 2023-06-29
Solution

diana-shyrokikh:


db/models.py
```diff
@@ -1 +1,39 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(
+        max_length=255,
+        unique=True
+    )
+    description = models.TextField(blank=True)
```
DmytroTsehelnyk, line 9: I think you should add choices here, since the player can only choose 1 of 4 races

DmytroTsehelnyk (approved):
GJ

main.py
```diff
@@ -1,10 +1,53 @@
...
+    for player, info in players.items():
+        race_name, race_description, skills = info["race"].values()
+
+        if not Race.objects.filter(name=race_name).exists():
+            Race.objects.create(
+                name=race_name,
+                description=race_description
+            )
+
+        player_race = Race.objects.get(name=race_name)
```
YuriiKindrat, line 21: try to use get_or_create method to simplify ur code

YuriiKindrat: the same is for simillar cases

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job

PR #398, opened 2023-06-29
Solution for Game Models

OlhaTryhub:


main.py
```diff
@@ -1,10 +1,47 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_file:
+        players = json.load(players_file)
+
+    for nickname_, properties in players.items():
+        if not Race.objects.all().filter(name=properties["race"]["name"]):
```
diana-shyrokikh, line 12: You can also use  Model.objects.filter(some_field="data").exists() (it is the hint from README) to check if we already have the object in database

diana-shyrokikh (approved):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
YuriiKindrat, line 26: did u read checklist ?

main.py
YuriiKindrat: agree, also no need to use all method before filter )
try to use get_or_create method to simplify ur code

YuriiKindrat (changes requested):

Oleg-s-s (approved):
Good job

PR #400, opened 2023-06-29
solution py-game-models

mariiahorbova:


main.py
```diff
@@ -1,10 +1,44 @@
...
+    for player, player_info in players_info.items():
+        guild = player_info["guild"] if player_info["guild"] else None
+        if guild:
+            guild, _ = Guild.objects.get_or_create(
+                name=guild["name"],
+                description=guild["description"]
+            )
+
+        race = player_info["race"]
+        if race:
```
oleksiikolii, line 21: You don't need this check; race always exists and even if it is not table in the db will not accept you create row without race or null in this column.

oleksiikolii (approved):

YuriiKindrat (approved):

main.py
```diff
@@ -1,10 +1,43 @@
...
+                name=skill["name"],
+                bonus=skill["bonus"],
+                race=Race.objects.get(name=race["name"])
+            )
+
+        Player.objects.create(
+            nickname=player,
+            email=player_info["email"],
+            bio=player_info["bio"],
+            race=Race.objects.get(name=race["name"]),
```
YuriiKindrat, line 38: could u create a race variable on 21 line, to use it here?

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🐞

PR #402, opened 2023-06-30
Solution #1

erikagayan:


main.py
```diff
@@ -1,10 +1,46 @@
...
+        else:
+            guild = None
+
+        Player.objects.create(
+            nickname=player_name,
+            email=player_info["email"],
+            bio=player_info["bio"],
+            race=race,
+            guild=guild,
+        )
```
RedCatHot, line 43: Everything is okay, but if u wanna, u can try to split this into functions
Maybe mentor will ask u for that

RedCatHot (changes requested):
U forgot to add migrations, cause of that tests failed. U can do this even here, just add file into git project, and I'll aprove this

RedCatHot (approved):

main.py
```diff
@@ -1,10 +1,46 @@
...
+        race = player_info["race"]
+        guild = player_info["guild"]
+
+        race, is_race_created = Race.objects.get_or_create(
+            name=race["name"],
+            description=race["description"]
+        )
+
+        if is_race_created:
+            for skill in player_info["race"]["skills"]:
```
LiudmylaKulinchenko, line 22: You will get an error if skill is already exists, you should check it

LiudmylaKulinchenko: fix in guild, race and user

LiudmylaKulinchenko (changes requested):

main.py
LiudmylaKulinchenko: Don't forget to use `race` variable here

Oleg-s-s (changes requested):
It would be better to split main.py into functions to make you code more modular

main.py
```diff
@@ -1,10 +1,58 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
+def create_races(players: dict) -> None:
```
LiudmylaKulinchenko, line 8: consider to change function names. You not just create races, but also skills. Same in the next func

LiudmylaKulinchenko (changes requested):
Well done

YuriiKindrat (approved):

PR #403, opened 2023-06-30
'Solution'

RedCatHot:


Tarasidze (changes requested):
add migrations pls
everything else is OK!

main.py
```diff
@@ -1,10 +1,45 @@
...
+        else:
+            guild = None
+
+        Player.objects.create(
+            nickname=player,
+            email=player_info["email"],
+            bio=player_info["bio"],
+            race=race,
+            guild=guild,
+        )
```
Oleg-s-s, line 42: Split this into functions

Oleg-s-s (changes requested):

main.py
```diff
@@ -1,10 +1,73 @@
...
+
+
 def main() -> None:
-    pass
+    players = load_players_from_file()
+
+    for player, player_info in players.items():
+        race_info = player_info.get("race")
+        skills_info = race_info.get("skills")
+        guild_info = player_info.get("guild")
```
YuriiKindrat, line 64: no need to create additional variables here

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
YuriiKindrat, line 24: it seems to me, that race could not be a null
image

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job 🎈

PR #409, opened 2023-07-01
'Solution'

yakhlamova:


db/models.py
```diff
@@ -1 +1,32 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    create_at = models.DateTimeField(auto_now=True)
```
Altahoma, line 32: From checklist:
Don't confuse `auto_now` and `auto_now_add`
- When you use `auto_now_add` - it records the time and date of creation.
- When you use `auto_now` - updates the value of the field to the current time and date every time when the object changes.

Altahoma (changes requested):

db/models.py
yakhlamova: In the task we need: `created_at - a DateTime field, that is set with the **current time** by default`, or I misunderstood?

Altahoma (approved):

db/models.py
LiudmylaKulinchenko: When you'll update a user, `created_at` value will be changed if you are using auto_now
But with auto_now_add it won't be changed
Is it clear for you now? 😊

main.py
```diff
@@ -1,10 +1,47 @@
 import init_django_orm  # noqa: F401
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
```
LiudmylaKulinchenko, line 7: split into functions to make more modular

LiudmylaKulinchenko (changes requested):

main.py
```diff
@@ -1,10 +1,65 @@
...
 
+def create_race(race: dict) -> None:
+    if not Race.objects.filter(name=race["name"]).exists():
+        Race.objects.create(
+            name=race["name"],
+            description=race["description"]
+        )
+
+
+def get_race_id(race: dict) -> int:
```
YuriiKindrat, line 15: bad func naming and type annotation 
are u sure that get method returns an int ?

main.py
```diff
@@ -1,10 +1,65 @@
...
+def create_guild(guild: dict) -> None:
+    if guild and not Guild.objects.filter(name=guild["name"]).exists():
+        Guild.objects.create(
+            name=guild["name"],
+            description=guild["description"]
+        )
+
+
+def get_guild_id(guild: dict) -> int:
+    return Guild.objects.get(name=guild["name"])
```
YuriiKindrat, line 38: same is here

main.py
```diff
@@ -1,10 +1,65 @@
...
 def main() -> None:
-    pass
+    with open("players.json") as data_file:
+        data = json.load(data_file)
+
+    for player in data:
+        player_data = data[player]
+        race = player_data["race"]
+        skills = race["skills"]
+        guild = player_data["guild"]
```
YuriiKindrat, line 49: no need to create additional variables here

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🐇

PR #412, opened 2023-07-02
Solution

volodymyr-komarnyckyi:


kovaliskoveronika (approved):

main.py
```diff
@@ -1,10 +1,43 @@
...
+                name=guild["name"],
+                description=guild["description"]
+            )
+
+        race = player_info["race"]
+        Race.objects.get_or_create(
+            name=race["name"],
+            description=race["description"]
+        )
+        racing = Race.objects.get(name=race["name"])
```
YuriiKindrat, line 24: could u create this variable using get_or_create method before ?) why do u need to make query to db twice ?

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
```
LiudmylaKulinchenko, line 5: You can also add `choices` here

PR #423, opened 2023-07-14
Solution

alecilchenko:


main.py
```diff
@@ -1,10 +1,54 @@
...
+            if not Guild.objects.filter(name=guild["name"]).exists():
+                obj = Guild(
+                    name=value["guild"]["name"],
+                    description=value["guild"]["description"]
+                )
+                obj.save()
+                value["guild"] = obj
+            else:
+                obj = Guild.objects.get(name=guild["name"])
+                value["guild"] = obj
```
pavlejviki, line 23: To simplify this block of code you can use get_or_create() method.

main.py
```diff
@@ -1,10 +1,54 @@
...
+        if race["skills"]:
+            for skill in race["skills"]:
+                if not Skill.objects.filter(name=skill["name"]).exists():
+                    obj = Skill(
+                        name=skill["name"],
+                        bonus=skill["bonus"],
+                        race=race_obj
+                    )
+                    obj.save()
+    for name, value in data.items():
```
pavlejviki, line 43: this for loop is redundant, you can create the Player object inside the previous loop and you need to assign to the race and guild attributes of the Player  the corresponding objects from the related model, rather than just their names.

main.py
```diff
@@ -1,10 +1,54 @@
...
+        if not Race.objects.filter(name=race["name"]).exists():
+            race_obj = Race(
+                name=race["name"],
+                description=race["description"]
+            )
+            race_obj.save()
+            race["name"] = race_obj
+        else:
+            race_obj = Race.objects.get(name=race["name"])
+            race["name"] = race_obj
```
pavlejviki, line 33: use get_or_create() method

main.py
```diff
@@ -1,10 +1,54 @@
...
+        if race["skills"]:
+            for skill in race["skills"]:
+                if not Skill.objects.filter(name=skill["name"]).exists():
+                    obj = Skill(
+                        name=skill["name"],
+                        bonus=skill["bonus"],
+                        race=race_obj
+                    )
```
pavlejviki, line 41: use get_or_create() method

pavlejviki (changes requested):

Bohdan-Salamakha (approved):

PR #424, opened 2023-07-15
Solution

BogdanFSD:


main.py
```diff
@@ -1,10 +1,58 @@
...
+        name=race_data.get("name"),
+        description=race_data.get("description")
+    )
+    return race
+
+
+def create_skills(skills_data: dict, race: Race) -> None:
+    for skill in skills_data:
+        if not Skill.objects.filter(name=skill.get("name")).exists():
+            Skill.objects.create(
```
ArsenPidhoretskyi, line 18: You can use get_or_create
Fix in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #426, opened 2023-07-18
Solution

Davison322:


main.py
```diff
@@ -1,10 +1,46 @@
...
+        email = player_info.get("email")
+        bio = player_info.get("bio")
+        race_ = player_info.get("race")
+        skills_ = race_.get("skills")
+        guild_ = player_info.get("guild")
+
+        race = Race.objects.get_or_create(
+            name=race_.get("name"),
+            description=race_.get("description")
+        )[0]
```
ArsenPidhoretskyi, line 20:
```suggestion
        race, _ = Race.objects.get_or_create(
            name=race_.get("name"),
            description=race_.get("description")
        )
```
Better to do like this)
Fix in all places

main.py
```diff
@@ -1,10 +1,46 @@
...
+            )[0]
+        else:
+            guild = None
+
+        Player.objects.create(
+            nickname=player_nickname,
+            email=email,
+            bio=bio,
+            guild=guild,
+            race_id=race.id
```
ArsenPidhoretskyi, line 42:
```suggestion
            race=race
```

ArsenPidhoretskyi (changes requested):

main.py
Davison322: Bu why its better?

ArsenPidhoretskyi: It is more clear code. If you know that it will return tuple with two params better to use: 
variable, _
In all future project you will use this construction instead of [0]
[0] you can use if there unknown count of parameters

Grayder0152 (approved):

PR #427, opened 2023-07-18
- solution

Oleksiy-Lyashenko:


main.py
```diff
@@ -1,10 +1,61 @@
...
+                )
+                guild.save()
+
+        race = Race.objects.get(name=info_race["name"])
+        guild = None
+
+        if info_guild is None:
+            guild = None
+        else:
+            guild = Guild.objects.get(name=info_guild["name"])
```
Dimosphen1, line 49: Add this operation to the first `if` condition related to the `info_guild`.

main.py
```diff
@@ -1,10 +1,61 @@
...
+                    Skill.objects.filter(name=skill_detail["name"]).exists()
+                ):
+                    skill = Skill(
+                        name=skill_detail["name"],
+                        bonus=skill_detail["bonus"],
+                        race=race,
+                    )
+                    skill.save()
+
+        if info_guild is not None:
```
Dimosphen1, line 35: You can shorten this check to `if info_guild:`, it would have enough explicitly.

Dimosphen1 (changes requested):

db/models.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "It stores a short description provided by a user",
+        max_length=255
+    )
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
pavlejviki, line 30: - Don't confuse the functionality of the on_delete parameter

Use on_delete=models.CASCADE if you know that some model doesn't exist without other model. So when the referenced object is deleted, also delete the objects that have references to it.

Use on_delete=models.SET_NULL if you know that the object should not be deleted, even if the object it references may be deleted.

So in this case SET_NULL will be preferable.

main.py
```diff
@@ -1,10 +1,60 @@
...
-    pass
+    with open("players.json", "r") as players_json:
+        list_of_players = json.load(players_json)
+
+    for name, info in list_of_players.items():
+        info_race = info["race"]
+        list_of_info_skills = info["race"]["skills"]
+        info_guild = info["guild"]
+
+        race = info_race["name"]
```
pavlejviki, line 16: You used the same variable name, as for the newly created Race object. It should be different.

main.py
```diff
@@ -1,10 +1,60 @@
...
+
+        if not Race.objects.filter(name=race).exists():
+            race = Race(
+                name=race,
+                description=info_race["description"]
+            )
+            race.save()
+        else:
+            race = Race.objects.get(name=race)
+
```
pavlejviki, line 27: Though your code works correct, you can simplify it using get_or_create() function.

pavlejviki (changes requested):

PR #431, opened 2023-07-20
'Solution'

Jorjio22:


main.py
```diff
@@ -1,10 +1,43 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+
+    for player_data in data.values():
+        race_data = player_data["race"]
```
olena-hul, line 12: Use .get() method instead of the index access to avoid KeyError
Fix everywhere

olena-hul (changes requested):

EdAlekseiev (approved):
Good job

PR #439, opened 2023-08-02
Solution game-models

MarkShapka:


main.py
```diff
@@ -1,10 +1,55 @@
...
+                        name=skill.get("name"),
+                        bonus=skill.get("bonus"),
+                        race=race_from_db
+                    )
+
+        new_player.race = race_from_db
+        if not Player.objects.filter(nickname=new_player.nickname).exists():
+            new_player.save()
+        print(new_player)
+        print("---------------")
```
Dimosphen1, line 52: Delete these prints, they are not needed.

main.py
```diff
@@ -1,10 +1,55 @@
...
+
+        guild = data.get("guild")
+        if guild is not None:
+            guild_name = guild.get("name")
+            if not Guild.objects.filter(name=guild_name).exists():
+                Guild.objects.create(
+                    name=guild_name,
+                    description=guild.get("description")
+                )
+            guild_from_db = Guild.objects.get(name=guild_name)
```
Dimosphen1, line 26: There is no need to query the database additionally in case you create a record, `.create()` will return the required object. You need to write it to a variable. Execute `.get()` in the `else` statement to achieve that.

main.py
```diff
@@ -1,10 +1,55 @@
...
+            new_player.guild = guild_from_db
+
+        race_ = data.get("race")
+        race_name = race_.get("name")
+        if not Race.objects.filter(name=race_name).exists():
+            Race.objects.create(
+                name=race_name,
+                description=race_.get("description")
+            )
+        race_from_db = Race.objects.get(name=race_name)
```
Dimosphen1, line 36: There is no need to query the database additionally in case you create a record, `.create()` will return the required object. You need to write it to a variable. Execute `.get()` in the `else` statement to achieve that.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #442, opened 2023-08-07
solution

vkleshko:


main.py
```diff
@@ -1,10 +1,49 @@
...
+        guild_info = player_info.get("guild")
+        skills_info = race_info.get("skills") if race_info else None
+
+        if race_info:
+            race, _ = Race.objects.get_or_create(
+                name=race_info["name"],
+                description=race_info["description"]
+            )
+        else:
+            race = None
```
ArsenPidhoretskyi, line 22:
```suggestion
        race = None
        if race_info:
            race, _ = Race.objects.get_or_create(
                name=race_info["name"],
                description=race_info["description"]
            )
```
More better to read instead of if-else

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #454, opened 2023-08-23
Solution

pnakongit:


main.py
```diff
@@ -1,10 +1,50 @@
...
+
+        race_data = player_data["race"]
+        if Race.objects.filter(name=race_data["name"]).exists():
+            race = Race.objects.get(name=race_data["name"])
+        else:
+            race = Race.objects.create(
+                name=race_data["name"],
+                description=(race_data["description"]
+                             if race_data["description"] else None)
+            )
```
vsmutok, line 22: you can use here [get_or_create](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create)

vsmutok (changes requested):

main.py
vsmutok: Fix in all places

vsmutok (changes requested):

main.py
pnakongit: Thx a lot! Done

ArsenPidhoretskyi (approved):
Well done!

PR #459, opened 2023-09-01
Solution

eduardhabryd:


db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(unique=True, max_length=255)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
Sofbog, line 24: Do we need ` on_delete=models.SET_NULL, null=True`

main.py
```diff
@@ -1,10 +1,61 @@
...
+        players = json.load(file)
+
+    for player, values in players.items():
+        nickname = player
+        email = values["email"]
+        bio = values["bio"]
+        race = values["race"]
+
+        race_name = race["name"]
+        race_description = race["description"]
```
Sofbog, line 23: is there a way to refactor this place

Sofbog (changes requested):

db/models.py
eduardhabryd: > Do we need ` on_delete=models.SET_NULL, null=True`

It seems not correct to delete player if a race was deleted, isn't it?

main.py
```diff
@@ -1,10 +1,51 @@
...
+        if guild:
+            guild_name = guild["name"]
+            guild_description = guild["description"]
+
+            guild_obj, _ = Guild.objects.get_or_create(
+                name=guild_name, description=guild_description
+            )
+
+        Player.objects.create(
+            nickname=nickname,
```
Sofbog, line 43: How to handle the situation when nickname=player_name not unique

main.py
```diff
@@ -1,10 +1,51 @@
...
+        )
+
+        skills = race["skills"]
+        for skill in skills:
+            name = skill["name"]
+            bonus = skill["bonus"]
+
+            Skill.objects.get_or_create(name=name, bonus=bonus, race=race_obj)
+
+        guild = values.get("guild")
```
Sofbog, line 31: read what get method will return if key does not exist, mb we can refactor this place

Sofbog (changes requested):

PR #462, opened 2023-09-04
Solution

tanya-burdenko:


main.py
```diff
@@ -1,10 +1,48 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        players = json.load(file)
+
+    for value in players.values():
+        if value["guild"] is not None:
+            if not Guild.objects.filter(name=value["guild"]["name"]).exists():
+                Guild.objects.create(
```
ArsenPidhoretskyi, line 15: You can use get_or_create
Fix in all places

main.py
```diff
@@ -1,10 +1,48 @@
...
+                    name=value["guild"]["name"],
+                    description=value["guild"]["description"]
+                )
+
+    for value in players.values():
```
ArsenPidhoretskyi, line 20: Why you need again loop?

ArsenPidhoretskyi (changes requested):
You can use 1 loop instead of 4 loops

main.py
```diff
@@ -1,10 +1,43 @@
...
+            Race.objects.get_or_create(
+                name=info["race"]["name"],
+                description=info["race"]["description"]
+            )
+
+            for skill in info["race"]["skills"]:
+                Skill.objects.get_or_create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=Race.objects.get(name=info["race"]["name"])
```
vasylhnatiuk, line 29: you already get race in 20 line, don't repeat yourself

main.py
```diff
@@ -1,10 +1,43 @@
...
+                )
+
+        Player.objects.create(
+            nickname=name,
+            email=info["email"],
+            bio=info["bio"],
+            race=Race.objects.get(name=info["race"]["name"]),
```
vasylhnatiuk, line 36: same

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):

PR #466, opened 2023-09-06
Solution

9rosLove:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
```
innyshka, line 5: name must be unique for the task

innyshka (approved):
Good job) Only add unique to name

PR #467, opened 2023-09-07
solution

sashabryl:


db/models.py
```diff
@@ -1 +1,38 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.CharField(max_length=255, blank=True)
+
+    def __str__(self) -> str:
+        return str(self.name)
```
pavlejviki, line 9: self.name is already a string

db/models.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
+
+    def __str__(self) -> str:
+        return str(self.name)
```
pavlejviki, line 18: self.name is already a string

db/models.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+    def __str__(self) -> str:
```
pavlejviki, line 25: self.name is already a string

main.py
```diff
@@ -1,10 +1,48 @@
...
+                    )
+        if info.get("guild"):
+            guild_d = info.get("guild")
+            if Guild.objects.filter(name=guild_d.get("name")).exists():
+                guild = Guild.objects.get(name=guild_d.get("name"))
+            else:
+                guild = Guild.objects.create(
+                    name=guild_d.get("name"),
+                    description=guild_d.get("description")
+                )
```
pavlejviki, line 35: you can simplify your code using get_or_create() method

pavlejviki (changes requested):

PR #468, opened 2023-09-07
'Solution'

lgSilay:


OlegatorLE (approved):
Looks good, but u canread about get_or_create() method and simplify your code

lgSilay:
> Looks good, but u canread about get_or_create() method and simplify your code

Thank for this advise!)

PR #470, opened 2023-09-07
implemented solution

innyshka:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
lgSilay, line 26: probably better to use auto_now_add instead auto_now.
Info from checklist:
"When you use auto_now_add - it records the time and date of creation.

When you use auto_now - updates the value of the field to the current time and date every time when the object changes."

main.py
```diff
@@ -1,10 +1,48 @@
 import init_django_orm  # noqa: F401
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+
+    with open("players.json", "r") as file:
```
lgSilay, line 9: flag "r" set as default, you can don't use them, if you want just to read file information

lgSilay (approved):
Good job!) Check my small comments)

db/models.py
innyshka: thanks)

Sofbog (changes requested):

PR #472, opened 2023-09-07
solution

arinaoltarzhevska:


main.py
```diff
@@ -1,10 +1,59 @@
...
+        race_name = race_data["name"]
+        race_description = race_data["description"]
+
+        if not Race.objects.filter(name=race_name).exists():
+            race_obj = Race.objects.create(
+                name=race_name,
+                description=race_description
+            )
+        else:
+            race_obj = Race.objects.get(name=race_name)
```
ArsenPidhoretskyi, line 26: You can use get_or_create
Fix in all places. Instead of use create + get + exists you can just use get_or_create

ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,64 @@
...
+        #         description=race_description
+        #     )
+        # else:
+        #     race_obj = Race.objects.get(name=race_name)
+
+        for skill in race_data["skills"]:
+            skill_name = skill["name"]
+            skill_bonus = skill["bonus"]
+            if not Skill.objects.filter(name=skill_name).exists():
+                Skill.objects.create(
```
ArsenPidhoretskyi, line 37: Use get or create

main.py
```diff
@@ -1,10 +1,64 @@
...
+        if guild_data := player_data["guild"]:
+            guild_name = guild_data["name"]
+            if not Guild.objects.filter(name=guild_name).exists():
+                guild_description = guild_data["description"]
+                guild_obj = Guild.objects.create(
+                    name=guild_name,
+                    description=guild_description
+                )
+            else:
+                guild_obj = Guild.objects.get(name=guild_name)
```
ArsenPidhoretskyi, line 53: Use get or create

ArsenPidhoretskyi (changes requested):

PR #473, opened 2023-09-07
Develop

Barsh4ec:


KristianYe (approved):
LGTM!

main.py
```diff
@@ -1,10 +1,40 @@
...
+        if skills := player_data["race"]["skills"]:
+            for skill in skills:
+                Skill.objects.get_or_create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=player_race
+                )
+
+        guild = player_data["guild"]
+        if guild is not None:
```
Sofbog, line 26: `if guild:` will make the same check as `if guild is not None:`

Sofbog (changes requested):

PR #474, opened 2023-09-07
Implemented game models and players creating

KristianYe:


9rosLove (approved):

db/models.py
```diff
@@ -1 +1,42 @@
...
+        unique=True,
+        max_length=255
+    )
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(unique=True, max_length=255)
+    bonus = models.CharField(
+        "Bonus describes what kind of bonus player can get from skill",
```
Sofbog, line 22: read about Pros and Cons before add description, its better to name field more clearly

db/models.py
```diff
@@ -1 +1,42 @@
...
+        ("elf", "Elf"),
+        ("dwarf", "Dwarf"),
+        ("human", "Human"),
+        ("ork", "Ork")
+    )
+    name = models.CharField(
+        choices=RACE_NAME_CHOICES,
+        unique=True,
+        max_length=255
+    )
```
Sofbog, line 15: what will happen when you will need to add more choices ?

db/models.py
```diff
@@ -1 +1,42 @@
...
+
+class Guild(models.Model):
+    name = models.CharField(unique=True, max_length=255)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
```
Sofbog, line 36: read about Pros and Cons before add description, its better to name field more clearly

main.py
```diff
@@ -1,10 +1,52 @@
...
+            player_race = Race.objects.create(
+                name=race["name"],
+                description=race.get("description", "")
+            )
+            for skill in race["skills"]:
+                Skill.objects.create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=player_race
+                )
```
Sofbog, line 30: read about `get_or_create` method

main.py
```diff
@@ -1,10 +1,52 @@
...
+
+        if guild:
+            if Guild.objects.filter(name=guild["name"]).exists():
+                player_guild = Guild.objects.get(name=guild["name"])
+            else:
+                player_guild = Guild.objects.create(
+                    name=guild["name"],
+                    description=guild.get("description")
+                )
```
Sofbog, line 39: read about `get_or_create` method

Sofbog (changes requested):

db/models.py
KristianYe: We can simply add the necessary race name to choices, if we'll get an update. In our task has written, that we have 4 races:
image
I think it would be better to add the available races to the choices to avoid duplicate races that would only differ by 1 letter due to a missclick and add new races to choices if it'll be necessary.

Sofbog (approved):
Well Done

db/models.py
Sofbog: `I think it would be better to add the available races to the choices to avoid duplicate races that would only differ by 1 letter due to a missclick and add new races to choices if it'll be necessary.`
agree

PR #475, opened 2023-09-08
Implemented game models and added players to DB

MarianKovalyshyn:


main.py
```diff
@@ -1,10 +1,56 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json") as source_file:
+        players_data = json.load(source_file)
+
+    for player in players_data:
```
denys-source, line 12: I think it is better to unpack player's nickname and data into two variables

main.py
```diff
@@ -1,10 +1,56 @@
...
+        player_guild = players_data[player]["guild"]
+        if player_guild:
+            if not Guild.objects.filter(name=player_guild["name"]).exists():
+                guild_ = Guild(
+                    name=player_guild["name"],
+                    description=player_guild["description"],
+                )
+                guild_.save()
+            else:
+                guild_ = Guild.objects.filter(name=player_guild["name"])[0]
```
denys-source, line 23: it is better to use `get` here

main.py
```diff
@@ -1,10 +1,56 @@
...
-    pass
+    with open("players.json") as source_file:
+        players_data = json.load(source_file)
+
+    for player in players_data:
+        guild_ = None
+        player_guild = players_data[player]["guild"]
+        if player_guild:
+            if not Guild.objects.filter(name=player_guild["name"]).exists():
+                guild_ = Guild(
```
denys-source, line 17: In my opinion, using orm instead of class instances and `save` is better

main.py
```diff
@@ -1,10 +1,56 @@
...
+                    name=player_guild["name"],
+                    description=player_guild["description"],
+                )
+                guild_.save()
+            else:
+                guild_ = Guild.objects.filter(name=player_guild["name"])[0]
+
+        race_ = None
+        player_race = players_data[player]["race"]
```
denys-source, line 26: As for me, these variable names are not clear enough

ochernous (approved):
Good job!

For simplification and improvement and better readability of the code, you can use Denis's suggestions.

main.py
MarianKovalyshyn: How would you name these variables?

main.py
MarianKovalyshyn: In the task there is a sentence 'Create only one instance for each guild, race and skill, do not copy them'. So I thought we needed to use class instances for that

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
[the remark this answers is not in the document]
MarianKovalyshyn: I don't understand what is wrong with my code. I think that Player can't be without Race, Player can be without guild and auto_now is used to set current time by default. I will change 'auto_now' to 'auto_now_add' but I am not sure about this, so please give more info about my mistake.

Sofbog (approved):
LGTM

PR #492, opened 2023-09-12
Solution

obohatov:


obohatov:
Thank you for advice try to use get_or_create method!

main.py
```diff
@@ -1,10 +1,46 @@
...
+                description=guild["description"]
+            )
+
+        Player.objects.create(
+            nickname=player_name,
+            email=player_data["email"],
+            bio=player_data["bio"],
+            race=race,
+            guild=guild
+        )
```
Sofbog, line 43: How to handle the situation when `nickname=player_name` not unique

Sofbog (changes requested):

PR #494, opened 2023-09-13
"Solution"

WonnaKoo:


db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(unique=True, max_length=255)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
Sofbog, line 24: do we really need  `on_delete=models.SET_NULL` here?

main.py
```diff
@@ -1,10 +1,39 @@
...
+                name=player_data["race"]["name"],
+                description=player_data["race"]["description"]
+            )
+        for player_skill in player_data["race"]["skills"]:
+            Skill.objects.get_or_create(
+                name=player_skill["name"],
+                bonus=player_skill["bonus"],
+                race=race
+            )
+        guild = player_data["guild"]
```
Sofbog, line 24: what if there is no data in `player_data["guild"]`

Sofbog (changes requested):

PR #499, opened 2023-09-21
Solution

AlexPelevin99:


db/models.py
```diff
@@ -1 +1,37 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        to=Race,
+        on_delete=models.CASCADE,
```
Dimosphen1, line 14: Do not forget to add `related_name` when you use ForeignKey in future tasks.

db/models.py
```diff
@@ -1 +1,37 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
```
Dimosphen1, line 6: It would be better to always define the `__str__` method when creating a new model.

Dimosphen1 (approved):
LGTM, well done! Several comments are left for consideration.

PR #501, opened 2023-09-21
Develop

Daniil-Pankieiev:


main.py
```diff
@@ -2,9 +2,57 @@
...
-    pass
+    try:
+        with open("players.json") as file:
+            data = json.load(file)
+    except FileNotFoundError:
+        print("File 'players.json' not found.")
+        return
+    except Exception as e:
+        print(f"An error occurred: {e}")
+        return
```
LiudmylaKulinchenko, line 17: Why don't you want to let those errors raise?

LiudmylaKulinchenko (changes requested):

main.py
Daniil-Pankieiev: I thought it was a good idea

LiudmylaKulinchenko: It's fine for FileNotFound, but you should not catch all errors using Exception

PR #512, opened 2023-10-07
Solution

MaksymUK:


db/models.py
```diff
@@ -1 +1,42 @@
...
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(
+        "what kind of bonus players can get", max_length=255,
+    )
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
ArsenPidhoretskyi, line 20: Why do you need f-string here?

db/models.py
```diff
@@ -1 +1,42 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
ArsenPidhoretskyi, line 28: Why do you need f-string here?

db/models.py
```diff
@@ -1 +1,42 @@
...
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "short description provided by a user about himself/herself",
+        max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self) -> str:
+        return f"{self.nickname}"
```
ArsenPidhoretskyi, line 42: Why do you need f-string here?

ArsenPidhoretskyi (changes requested):

db/models.py
MaksymUK: Because if I write self.name only, it shows warning - Expected type 'str', got 'CharField' instead

ArsenPidhoretskyi: It is PyCharm warning, not python warning

ArsenPidhoretskyi (approved):
Well done!

PR #516, opened 2023-10-10
'Solution'

hanovskyiv:


main.py
```diff
@@ -1,10 +1,37 @@
...
+            Skill.objects.get_or_create(
+                name=skill["name"],
+                bonus=skill["bonus"],
+                race=race
+            )
+
+        guild, _ = Guild.objects.get_or_create(
+            name=data["guild"]["name"],
+            description=data["guild"]["description"]
+        ) if data["guild"] else (None, _)
```
viktoria-rybenchuk, line 26: Why do you need  the underscore `_`

viktoria-rybenchuk (changes requested):

main.py
hanovskyiv: When using get_or_create we get a tuple back, also If the player is not a member of a guild we also need to return it to a tuple,
If delete underscore we get error.

Polyakiv-Andrey (changes requested):

db/models.py
```diff
@@ -1 +1,39 @@
 from django.db import models
+from django.db.models import CharField
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField()
+
+    def __str__(self) -> CharField:
```
viktoria-rybenchuk, line 9: it just return `str`. Fix in all places

viktoria-rybenchuk (changes requested):
remove solution.py from your repository

viktoria-rybenchuk (approved):
GJ!

PR #517, opened 2023-10-12
Develop

MarinaKhrunina:


main.py
```diff
@@ -1,10 +1,46 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players_data = json.load(file)
+
+    for player, data in players_data.items():
+        race_info = Race.objects.get_or_create(
```
ArsenPidhoretskyi, line 13: Better to do like this:
```suggestion
        race, _ = Race.objects.get_or_create(
```
The same in other places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #531, opened 2023-10-26
'Solution'

lylakivan:


db/models.py
```diff
@@ -1 +1,32 @@
...
+
+
+class Race(models.Model):
+    RACE_CHOICES = (
+        ("Elf", "Elf"),
+        ("Dwarf", "Dwarf"),
+        ("Human", "Human"),
+        ("Ork", "Ork"),
+    )
+    name = models.CharField(max_length=255, choices=RACE_CHOICES)
```
pavlejviki, line 11: there is no need to limit the name field to certain choices

pavlejviki (changes requested):

db/models.py
```diff
@@ -1 +1,32 @@
 from django.db import models
+
+
+class Race(models.Model):
+    RACE_CHOICES = (
+        ("Elf", "Elf"),
```
pavlejviki, line 6: you can remove RACE_CHOICES as well

pavlejviki (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
```
Polyakiv-Andrey, line 20: add str method

Polyakiv-Andrey (changes requested):

db/models.py
```diff
@@ -1 +1,38 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
+    description = models.TextField(blank=True)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
viktoria-rybenchuk, line 9: why do you need `f-string`?

db/models.py
```diff
@@ -1 +1,38 @@
...
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self) -> str:
+        return f"{self.nickname}"
```
viktoria-rybenchuk, line 38: why do you need f-string?

viktoria-rybenchuk (changes requested):

db/models.py
viktoria-rybenchuk: i mean you can simple return self.nickname. Don't remove `str` method

viktoria-rybenchuk (changes requested):
Don't remove `str` method

db/models.py
```diff
@@ -1 +1,39 @@
...
+
+    def __str__(self) -> CharField:
+        return self.name
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+    def __str__(self) -> CharField:
```
Dimosphen1, line 26: Although it could raise warnings in different IDEs, it would be more preferable to indicate that the return type is `str` for the `__str__` method, because it would be a value that is eventually returned, not an instance of the `CharField`. Wrapping the `self.name` in f-string could be a workaround for this issue, but it leads to an unnecessary line formatting that can be omitted. For this specific case, ignore warnings in the IDE, indicate that the return type is `str`, and leave the return as it is with a single value. It relates only to cases when you return a single value for a single `CharField` when you combine different fields in the `return` using f-string.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #532, opened 2023-10-26
task completed (1)

DaniilHinzburh:


main.py
```diff
@@ -1,10 +1,56 @@
...
+        guild_data = player_data.get("guild")
+
+        try:
+            race = Race.objects.get(name=race_data["name"])
+        except Race.DoesNotExist:
+            race = Race(
+                name=race_data["name"],
+                description=race_data.get("description")
+            )
+            race.save()
```
ArsenPidhoretskyi, line 21: Use get_or_create
Fix in all places instead of usage try-except

ArsenPidhoretskyi (changes requested):

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
...
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Guild',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
```
vasylhnatiuk, line 18: change ' to ", use Black

vasylhnatiuk (changes requested):

db/migrations/0001_initial.py
DaniilHinzburh: do you want me to fix the double quotes? but this code is generated automatically by the command python manage.py makemigrations db
I don't understand the point of editing something

vasylhnatiuk: it is good practice to use the same quotes for all of code lines

vasylhnatiuk (approved):

PR #540, opened 2023-10-31
Develop

prochigor:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
Familenko, line 26: ```When you use auto_now - updates the value of the field to the current time and date every time when the object changes.```
I think your hero will change date of creation in case of develop new skill or go to guild

main.py
```diff
@@ -1,10 +1,70 @@
...
+    race_name = player["race"]["name"]
+    guild = player.get("guild")
+    guild_name = guild["name"] if guild else None
+    Player.objects.create(
+        nickname=player_name,
+        email=player["email"],
+        bio=player["bio"],
+        race=Race.objects.get(name=race_name),
+        guild=Guild.objects.get(name=guild_name) if guild_name else None
+    )
```
Familenko, line 55: A bit of a strange function, on the one hand a hint of reusability, on the other fixed values of the player["bio"] type

main.py
```diff
@@ -1,10 +1,70 @@
...
+
+    if guild:
+        guild_name = guild.get("name")
+        guild_description = guild.get("description")
+
+        if not Guild.objects.filter(name=guild_name).exists():
+            Guild.objects.create(
+                name=guild_name,
+                description=guild_description
+            )
```
Familenko, line 41: You have a lot of .exists() search which have complicity O(n) compare to  get_or_create with the time complexity is typically O(log n) or O(1) because of the indexing. Also there is three independent loop with O(n) each. With big data all this code will run to slow

Familenko (changes requested):

main.py
prochigor: it's because we work with the file, and when we have new json we expect the same names like "bio"

main.py
prochigor: it makes sense thanks

LiudmylaKulinchenko (approved):
Good job! ☀

PR #541, opened 2023-11-01
Solution

YegorT:


main.py
```diff
@@ -1,10 +1,41 @@
...
+
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players = json.load(file)
+
+    for player in players:
```
ArsenPidhoretskyi, line 12: Iterate through .items() because you always do players[player]

ArsenPidhoretskyi (changes requested):

db/models.py
```diff
@@ -1 +1,33 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        "It stores a short description "
+        "provided by a user about himself/herself.",
```
YukihiroSM, line 28: This parameter is being used as verbose_name. It's not the best choice for verbose name in your case. 
You can read about verbose names here https://www.geeksforgeeks.org/verbose_name-django-built-in-field-validation/
Or here: https://docs.djangoproject.com/en/4.2/ref/models/options/#verbose-name
Remove this documentation string everywhere, please.

YukihiroSM (changes requested):

YegorT:
forgive my inattention - tired in the evening.

PR #545, opened 2023-11-01
Develop

D1st3f:


kostomeister (approved):
graaaaaaape

db/models.py
```diff
@@ -1 +1,26 @@
...
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
LiudmylaKulinchenko, line 26: `auto_now_add` would be better here, use checklist

LiudmylaKulinchenko: As you had to do this: `a DateTime field, that is set with the current time by default`. The difference is:- when you use `auto_now_add` - it records the time and date of creation- when you use `auto_now` - updates the value of the field to the current time and date every time when the object changes

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🦋

PR #552, opened 2023-11-02
Solution

YegorT:


db/models.py
```diff
@@ -1 +1,32 @@
...
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(
+        max_length=255,
+        verbose_name="About me:"
+    )
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
NikOneZ1, line 32: It would be better to use auto_now_add instead of the auto_now, because auto_now automatically sets the field to now every time the object is saved, not created
```suggestion
    created_at = models.DateTimeField(auto_now_add=True)
```

main.py
```diff
@@ -1,10 +1,42 @@
...
-    pass
+    with open("players.json", "r") as file:
+        players = json.load(file)
+
+    for player, player_data in players.items():
+        race, created = Race.objects.get_or_create(
+            name=player_data["race"]["name"],
+            description=player_data["race"]["description"]
+        )
+        if player_data["guild"]:
```
NikOneZ1, line 18: It is recommended to use the .get() method instead of [] because it is more readable and safer if there are no objects in the dictionary with the key you want to retrieve.

main.py
NikOneZ1, line 1: It would be fine to define variable with the race to not repeat the code like player_data["race"]

NikOneZ1 (changes requested):

main.py
```diff
@@ -1,10 +1,44 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players = json.load(file)
+
+    for player, player_data in players.items():
+        player_race = player_data.get("race")
```
ArsenPidhoretskyi, line 13:
```suggestion
        player_race = player_data.get("race", {})
```

ArsenPidhoretskyi (approved):
Good job! But now potentially NoneType error

PR #553, opened 2023-11-02
Solution/Alex

AlexIzmashkin:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
NikOneZ1, line 26: It would be better to use auto_now_add instead of the auto_now, because auto_now automatically sets the field to now every time the object is saved, not created
```suggestion
    created_at = models.DateTimeField(auto_now_add=True)
```

main.py
NikOneZ1, line 1: It is recommended to use the .get() method instead of [] because it is more readable and safer if there are no objects in the dictionary with the key you want to retrieve.

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #554, opened 2023-11-02
Solution

Shatrovskyi:


DenPrislipskyi (approved):
You can try name=info.get("name") instead name=info["race"]["name"], but your solution also nice)))

db/models.py
```diff
@@ -1 +1,39 @@
...
+        return f"Guild name: {self.name}"
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
LiudmylaKulinchenko, line 35: use checklist, `auto_now_add` would be better here

main.py
```diff
@@ -1,10 +1,49 @@
...
+
+            if skills_data is not None:
+                for skill in skills_data:
+                    Skill.objects.get_or_create(
+                        name=skill["name"],
+                        bonus=skill["bonus"],
+                        race_id=race_obj.id
+                    )
+
+            guild_data = info["guild"] if info["guild"] else None
```
LiudmylaKulinchenko, line 29: u can just use .get() here instead

LiudmylaKulinchenko (changes requested):

main.py
```diff
@@ -1,10 +1,47 @@
...
-    pass
+    with open("players.json", "r") as players_file:
+        players_data = json.load(players_file)
+
+        for player_name, info in players_data.items():
+
+            race_obj, race = Race.objects.get_or_create(
+                name=info["race"]["name"],
+                description=info["race"]["description"]
+            )
```
LiudmylaKulinchenko, line 16: Don't put unnesessary code into a context manager. You also have this point in your checklist

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🦋

PR #555, opened 2023-11-02
django orm

DenPrislipskyi:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
Familenko, line 26: use ```auto_now_add``` - it records the time and date of creation

main.py
```diff
@@ -1,10 +1,45 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players = json.load(file)
+
+    for key, value in players.items():
+        race = value.get("race")
```
Familenko, line 13: use descriptive variables

main.py
```diff
@@ -1,10 +1,45 @@
...
+        if race:
+            race, bool_ = Race.objects.get_or_create(
+                name=race.get("name"),
```
Familenko, line 16: ```bool_``` is not used anywhere in the code, you can hide it by ```_```

Familenko (changes requested):

main.py
LiudmylaKulinchenko: I agree, fix that. Also avoid such naming, it is a bad practice

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🦋

PR #560, opened 2023-11-02
Game with models

SelFisher1488:


oleksandr-pupchenko (approved):
better use " instead of '
some lines are hard to read as they're too long

LiudmylaKulinchenko (approved):
Well done 🦋

PR #568, opened 2023-11-04
Solution game-models

mtiunov:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
```
NikOneZ1, line 5: Name should be unique as stated in the task

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255)
```
NikOneZ1, line 10: Name should be unique as stated in the task

db/models.py
```diff
@@ -1 +1,26 @@
...
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255)
```
NikOneZ1, line 16: Name should be unique as stated in the task

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255)
```
NikOneZ1, line 21: nickname should be unique as stated in the task

db/models.py
```diff
@@ -1 +1,26 @@
...
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, null=True, on_delete=models.SET_NULL)
+    created_at = models.DateTimeField(auto_now=True)
```
NikOneZ1, line 26: It would be better to use auto_now_add instead of auto_now because auto_now sets every time objects are saved, but you only need to set the value of this field when the object is created

main.py
```diff
@@ -1,10 +1,45 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
+with open("players.json", "r") as file:
+    data = json.load(file)
```
NikOneZ1, line 8: It would be better to move this code to the main function

main.py
NikOneZ1, line 1: Try to use .get method instead of [] to avoid KeyError and to do code more readable

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #571, opened 2023-11-05
Solution Game models

Oxbay:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateField(auto_now=True)
```
LiudmylaKulinchenko, line 26: use checklist, the last point there is about auto_now and auto_now_add

LiudmylaKulinchenko (changes requested):

db/models.py
Oxbay: Done

LiudmylaKulinchenko (approved):
Well done 🦋

PR #572, opened 2023-11-06
py-game-model

stamaksim:


db/models.py
```diff
@@ -1 +1,33 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255)
+    bonus = models.CharField(
+        "field describes what kind of bonus "
```
ArsenPidhoretskyi, line 12: In this case better to use help_text=

ArsenPidhoretskyi (approved):
Good job!

PR #588, opened 2023-11-25
Done game models

OlenaLis123:


main.py
```diff
@@ -1,10 +1,53 @@
...
+
+    for name, value_player in players_dict.items():
+        Race.objects.get_or_create(
+            name=value_player["race"]["name"],
+            description=value_player["race"]["description"]
+        )
+        race_ = Race.objects.get(
+            name=value_player["race"]["name"],
+            description=value_player["race"]["description"]
+        )
```
EdAlekseiev, line 22: `get_or_create` returns a created instance. So, you don't need the next code
`        race_ = Race.objects.get(
            name=value_player["race"]["name"],
            description=value_player["race"]["description"]
        )`
 Fix it here and in other place too

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #600, opened 2023-12-02
Solution

paashkovaaa:


NikOneZ1 (changes requested):
It would be better to use .get() instead of the []

EdAlekseiev (approved):
Good job

PR #609, opened 2023-12-13
Game models impl

krismva:


main.py
```diff
@@ -1,10 +1,38 @@
...
+            description=data["race"]["description"]
+        )
+        for skill in data["race"]["skills"]:
+            Skill.objects.get_or_create(
+                name=skill["name"],
+                bonus=skill["bonus"],
+                race=race
+            )
+        guild = None
+        if data["guild"]:
```
vasylhnatiuk, line 24:
```suggestion
        if data.get("guild"):
```

vasylhnatiuk: because if data doesn't have this key it will reaise keyerror

vasylhnatiuk (changes requested):

PR #613, opened 2023-12-18
Solution

aleksandra-prognimak:


main.py
```diff
@@ -1,10 +1,45 @@
...
-    pass
+    data = []
+
+    with open("players.json", "r") as file:
+        data = load(file)
+
+    for nickname, item in data.items():
+        race = item["race"]
+        guild = item["guild"]
+        skills = race["skills"]
```
LiudmylaKulinchenko, line 16: You should use get method here, because you will get a key error, and a check in 30th line will not work, for example

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done

PR #617, opened 2023-12-25
cleared task

BGBroth:


main.py
```diff
@@ -1,10 +1,52 @@
-import init_django_orm  # noqa: F401
+import init_django_orm  # noqa
 
-from db.models import Race, Skill, Player, Guild
+import json
+from db.models import Race, Skill, Guild, Player
+
+
+def create_race(data: dict) -> Race:
```
Dimosphen1, line 7: Your function name is ambiguous since it is not supposed to create skills. Rename it to `create_race_with_skills` for proper encapsulation.

main.py
```diff
@@ -1,10 +1,52 @@
...
+    race_name = data["name"]
+    race, created = Race.objects.get_or_create(
```
Dimosphen1, line 9: You could skip the `created` variable with `_` since it is not used in the logic.

Dimosphen1 (changes requested):

PR #630, opened 2024-01-19
solution

OlgaRozd:


db/models.py
```diff
@@ -1 +1,28 @@
+# db/models.py
+
 from django.db import models
+
```
Polyakiv-Andrey, line 4: add str methods for each models

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #649, opened 2024-02-06
Solution

NataliiaKucher:


main.py
```diff
@@ -1,10 +1,44 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+
+    for player_name, player_data in data.items():
+        race_data = player_data["race"]
```
vasylhnatiuk, line 12: it would be better to use get() here as well

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #653, opened 2024-02-08
Solution

msymonovych:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
```
aLEKS-e3, line 5: race should have a unique name

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True, blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
aLEKS-e3, line 26: I suppose you should set `auto_no_add` here

main.py
```diff
@@ -1,10 +1,41 @@
...
+
+        player, created = Player.objects.get_or_create(
+            nickname=nickname,
+            email=player["email"],
+            bio=player["bio"],
+            race=race,
+            guild=guild
+        )
+        if created:
+            player.save()
```
aLEKS-e3, line 38: I'm not sure, but is it necessary?

aLEKS-e3 (changes requested):

main.py
msymonovych: you're right)

PR #663, opened 2024-02-12
Solution

yulias97:


main.py
```diff
@@ -1,10 +1,39 @@
...
-    pass
+    with open("players.json") as players:
+        players_info = json.load(players)
+
+    for player, info in players_info.items():
+        race, created = Race.objects.get_or_create(
+            name=info["race"]["name"],
+            description=info["race"]["description"]
+        )
+        for skill in info["race"]["skills"]:
```
barabarinov, line 17: You should check if race is created and then create skills!

barabarinov (approved):

main.py
yulias97: thanks, will fix it

PR #667, opened 2024-02-12
Develop

oksankakot:


main.py
```diff
@@ -1,10 +1,40 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players = json.load(file)
+
+        for player, data in players.items():
```
mdubyna, line 12: image

mdubyna (changes requested):
Do not overload the context manager.

main.py
oksankakot: Thanks. Fixed that

PR #685, opened 2024-02-18
Solution

govorunenko:


data_from_players/get_player_data.py
```diff
@@ -0,0 +1,27 @@
+from db.models import Race, Guild
+
+
+def get_player_data(players: dict) -> list:
```
vasylhnatiuk, line 4: why do you need this?

create_funcs/create_player.py
```diff
@@ -0,0 +1,21 @@
...
+from db.models import Player
+from players_json import players_data
+
+
+def create_player() -> Player:
+    player_lst = get_player_data(players_data)
+    for player in player_lst:
+        print(player)
+        Player.objects.get_or_create(
+            nickname=player[0],
```
vasylhnatiuk, line 14: it is better to use key instead of index

vasylhnatiuk (changes requested):
it is better to use key instead of index to get data

data_from_players/get_player_data.py
govorunenko: I've combined get_player_data function with create_player function.

create_funcs/create_player.py
govorunenko: Used keys.

vasylhnatiuk (changes requested):
you use `for player in player` few times and it is bad, don't repeat code, especially loops. fix it

govorunenko:
I moved the functionality to the module "create_player.py" and used loop "for player in players:" for once.

EdAlekseiev (approved):
Good job

PR #705, opened 2024-03-05
Solution

Vityk-4W:


main.py
```diff
@@ -1,10 +1,42 @@
...
+        else:
+            player_guild = [None]
+        race = json_data[player]["race"]
+        player_race = Race.objects.get_or_create(
+            name=race["name"],
+            description=race["description"]
+        )
+        skills = race["skills"]
+        for skill in skills:
+            if skill["name"] not in Skill.objects.all().values_list("name"):
```
YukihiroSM, line 27: you don't need this condition if you are using get_or_create. It will create if skill doesn't exists, and get if skill exists

YukihiroSM (changes requested):
Else looks good to me!

YukihiroSM (approved):
Nicely done!

PR #706, opened 2024-03-06
Solution

AlexTsikhun:
I use different approaches for get Race name to practice and I needed to use this object (via flake8):
**race** = Race.objects.get_or_create(
            name=values["race"]["name"],
            description=values["race"]["description"]
        )

```
                race=Race.objects.get(name=values["race"]["name"])
            race=Race.objects.get(name=race[0].name),

```


ArsenPidhoretskyi (changes requested):

main.py
```diff
@@ -1,10 +1,39 @@
...
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        players_data = json.load(file)
+
+    for nickname, values in players_data.items():
+        race = Race.objects.get_or_create(
+            name=values["race"]["name"],
+            description=values["race"]["description"]
+        )[0]
```
Dimosphen1, line 16: As an alternative approach to avoid indexing, you could use unpacking
```suggestion
        race, _ = Race.objects.get_or_create(
            name=values["race"]["name"],
            description=values["race"]["description"]
        )
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #707, opened 2024-03-06
in process

Serhii-Hrynyshyn:
i need a help, what can i do for next step?


EdAlekseiev (changes requested):
Fix flake8 and test. If you have problem with this write to chat `py_chat` or visit QnA

EdAlekseiev (approved):
Good job

PR #711, opened 2024-03-10
10_03_24

SerheiZhus:


main.py
```diff
@@ -1,10 +1,42 @@
...
+
+            if player_info["guild"] is not None:
+                guild, created_bool = Guild.objects.get_or_create(
+                    name=player_info["guild"]["name"],
+                    defaults={"description": player_info["guild"]
+                              ["description"]},)
+            else:
+                guild = None
+
+            Player.objects.get_or_create(
```
vsmutok, line 31: no need to use `get_or_create` here
```suggestion
            Player.objects.create(
```

vsmutok (changes requested):

PR #712, opened 2024-03-11
Solution_GTA

TarasFirst:


db/migrations/0002_alter_player_guild.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

main.py
```diff
@@ -1,10 +1,36 @@
...
+            player, _ = Player.objects.get_or_create(
+                nickname=player_nick,
+                email=email,
+                bio=bio,
+                race=race,
+                guild=guild
+            )
 
 
 if __name__ == "__main__":
```
Dimosphen1, line 36: Do not perform these operations within the scope of the context manager as mentioned [in the checklist](https://github.com/TarasFirst/py-game-models/blob/develop/checklist.md#1-do-not-overload-the-context-manager)

main.py
```diff
@@ -1,10 +1,36 @@
...
+                skill, _ = Skill.objects.get_or_create(
+                    name=skill_data["name"],
+                    bonus=skill_data["bonus"],
+                    race=race
+                )
+            guild = None
+            if guild_data:
+                guild, _ = Guild.objects.get_or_create(**guild_data)
+
+            player, _ = Player.objects.get_or_create(
```
Dimosphen1, line 27: There is no need to write the `get_or_create` response to variables since it is not used anywhere. You could omit it
```suggestion
            Player.objects.get_or_create(
```

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #718, opened 2024-03-14
Solution

twonumberfortyfives:


db/models.py
```diff
@@ -1 +1,38 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
```
pavlopro, line 5: Please add `unique `parameter

pavlopro (changes requested):
Small change requested

vasylhnatiuk (approved):

PR #730, opened 2024-03-24
Solution

Bujhm99:


Oleksl888 (changes requested):

db/models.py
```diff
@@ -1 +1,30 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.DO_NOTHING,
```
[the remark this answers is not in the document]
Bujhm99: I think if we delete race it's not nececary, moreover, we should not to delete hole player. We can rewrite him in base race (we don't have it in current situation, so let it be Null).

db/models.py
```diff
@@ -1 +1,29 @@
...
+                             related_name="race_player")
```
Oleksl888, line 26: would make sense to set related name to `player`

Bujhm99:
61dbGcJXvCL _AC_UF894,1000_QL80_
bright picture as bookmark

db/models.py
```diff
@@ -1 +1,29 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE,
```
Oleksl888, line 12: we allow this field to be NULL - so here would be also more appropriate to use SET_NULL strategy.
CASCADE would be appropriate if this field could not be nullable

Oleksl888 (changes requested):

Bujhm99:
image
BBPB

db/models.py
```diff
@@ -1 +1,29 @@
...
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL,
+                             related_name="players", null=True)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL,
+                              related_name="players", null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
Oleksl888, line 29: auto_now_add would be more appropriate here - it will populate created_at field on record creation as opposed to auto_now that will update field on every write operation

main.py
```diff
@@ -1,10 +1,51 @@
...
+
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file_json:
+        data = json.load(file_json)
+    for nickname, player_data in data.items():
+        race, _ = Race.objects.filter(
```
Oleksl888, line 12: here and below must be simplified:
```
            name=data["race"]["name"],
            description=data["race"]["description"],
        )

Oleksl888: no need to filter here and then call get_or_create - the last method will return a record if there is a match by name/description if exists and create new record otherwise. filter is redundant

Oleksl888 (approved):
Approved but please check and fix the new comments

PR #733, opened 2024-03-27
'Solution'

KravchenkoSerhiii:


db/migrations/0002_alter_guild_description_alter_player_bio_and_more.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

main.py
```diff
@@ -1,10 +1,38 @@
...
+        race, _ = Race.objects.get_or_create(
+            name=player["race"]["name"],
+            description=player["race"]["description"])
+
+        for skill_data in player["race"]["skills"]:
+            skill, _ = Skill.objects.get_or_create(
+                name=skill_data["name"],
+                bonus=skill_data["bonus"], race=race)
+
+        if player["guild"] is not None:
```
Dimosphen1, line 22: You could simplify this check
```suggestion
        if player.get("guild"):
```

main.py
```diff
@@ -1,10 +1,38 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players:
+        players = json.load(players)
+
+    for p_name in players:
```
Dimosphen1, line 11: It would be better to use more descriptive variable name here
```suggestion
    for player_name in players:
```

Dimosphen1 (changes requested):

PR #745, opened 2024-04-04
done

antonver:


db/migrations/0003_alter_player_guild.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

main.py
```diff
@@ -1,10 +1,36 @@
 import init_django_orm  # noqa: F401
-
+from json import load
```
Dimosphen1, line 2: Add a blank line between different import groups

Dimosphen1, line 2: Prefer using `import json` and calling load as `json.load` to be more descriptive in the code

main.py
```diff
@@ -1,10 +1,36 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = load(file)
+    for key, value in data.items():
```
Dimosphen1, line 9: Use more descriptive variable names instead of `key` and `value`

main.py
```diff
@@ -1,10 +1,36 @@
...
+        race, created = Race.objects.get_or_create(
```
Dimosphen1, line 10: Since `created` is not used, omit it here and in the other places
```suggestion
        race, _ = Race.objects.get_or_create(
```

Dimosphen1 (changes requested):

main.py
antonver: Ok, but I thought that it is better not to import the whole module, and just import the method that is needed is better.

main.py
```diff
@@ -1,10 +1,36 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+    for nickname, ability in data.items():
+        race, _ = Race.objects.get_or_create(
+            name=ability["race"]["name"],
+            description=ability["race"].get("description")
+        )
```
pavlopro, line 13: Define 'guild' as None. Then, use the 'get' method to check its value.

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #750, opened 2024-04-12
Solution

mwellick:


main.py
```diff
@@ -1,10 +1,50 @@
...
+            race_description = data["race"]["description"]
+
+            race_info = Race.objects.get_or_create(
+                name=race_name,
+                description=race_description
+            )
+            print(race_info)
+            guild_data = data["guild"]
+            if guild_data is not None:
+                guild_name = guild_data["name"]
```
EdAlekseiev, line 23: https://github.com/mate-academy/py-game-models/blob/master/checklist.md#:~:text=1)%20Do%20not%20overload%20the%20context%20manager.

EdAlekseiev (approved):

main.py
mwellick: Will it be a  good practice if I implement,for example, a separate  fucnction  in main.py  which will return a loaded data from json file and function main()  will add this data  to db?

mwellick: main() will have the same realization but without overloading context manager

PR #764, opened 2024-04-19
Solution

DmytroHlazyrin:


db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=True)
+    bio = models.TextField(max_length=255, blank=True)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Oleksl888, line 24: related_name here ?

main.py
```diff
@@ -1,10 +1,43 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as files:
+        players_data = json.load(files)
+
+    for player, player_info in players_data.items():
+        race, created = Race.objects.get_or_create(
```
Oleksl888, line 13: created var is not used. What would be the right way to signify this?

main.py
```diff
@@ -1,10 +1,43 @@
...
+                race=race
+            )
+
+        if player_info["guild"] is not None:
+            guild, created = Guild.objects.get_or_create(
+                name=player_info["guild"]["name"],
+                description=player_info["guild"]["description"]
+            )
+        else:
+            guild = None
```
Oleksl888, line 31: you can move this line above line 25 to avoid using extra else statement

main.py
```diff
@@ -1,10 +1,43 @@
...
+
+        created_player = Player.objects.create(
+            nickname=player,
+            email=player_info["email"],
+            bio=player_info["bio"],
+            race=race,
+            guild=guild
+        )
+        created_player.save()
```
Oleksl888, line 40: no need to call .save() explicitly since you are using manager for creation

Oleksl888 (changes requested):

fsocie7y (approved):
GJ 👍

PR #765, opened 2024-04-19
Solution

SashaKisliy:


db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=True)
+    bio = models.CharField(max_length=255, blank=True)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Oleksl888, line 24: add related name here and below

main.py
```diff
@@ -1,10 +1,40 @@
 import init_django_orm  # noqa: F401
 
+import json
```
Oleksl888, line 3: check order of imports

Oleksl888 (changes requested):

db/models.py
Oleksl888: > I showed you on the last QnA that tests don't pass if you add a related_name

it only dont pass if you add on Skills model. Here on Player is ok

fsocie7y (changes requested):

Oleksl888 (approved):
Nice!

PR #768, opened 2024-04-19
game db solution

panicua:
I had to chage my code cause description of a task isn't clear, I'll let you decide:
NVIDIA_Share_KaC056EZoQ

It says that each player should choose a race to play, so those races should be available for a player in DB from the start (my opinion) but, if you create all the 4 races in advance, you receive an error(pytest) that checks DB and sees orc, dwarf races (no players have those races in our json file, so they weren't created). But.. don't you think that it's smart to create all 4 races before player can chose from them?
so our db after running a code MUST look like this:
races: ```elf, human``` (because only those races players had in json).
but I think it would be better like this: ```elf, human, orc, dwarf.``` (sadly tests fail in this case)

I hope you understood what I mean

### DIFF
--- a/main.py
+++ b/main.py
@@ -49,7 +49,7 @@ def create_skills(player_race: dict[str], race_instance: Race) -> None:
         )
 
 
-def create_guild(player_guild: dict[str]) -> Guild:
+def create_guild(player_guild: dict[str]) -> Guild | None:
     """
     Creating unique guild name if player has a guild,
     description(not required) and writes guild to DB
@@ -62,3 +62,7 @@ def create_guild(player_guild: dict[str]) -> Guild:
             description=player_guild["description"]
         )
     return guild
+
+
+if __name__ == "__main__":
+    main()


main.py
```diff
@@ -1,11 +1,64 @@
...
+    """
+    for skill in player_race["skills"]:
+        skill, _ = Skill.objects.get_or_create(
+            name=skill["name"],
+            bonus=skill["bonus"],
+            race=race_instance
+        )
 
 
-if __name__ == "__main__":
```
Oleksl888, line 10: you shouldn't remove this

main.py
```diff
@@ -1,11 +1,64 @@
...
-    main()
+def create_guild(player_guild: dict[str]) -> Guild:
```
Oleksl888, line 52: check return type here

Oleksl888 (changes requested):

Oleksl888:
> I had to chage my code cause description of a task isn't clear, I'll let you decide: NVIDIA_Share_KaC056EZoQ
> 
> It says that each player should choose a race to play, so those races should be available for a player in DB from the start (my opinion) but, if you create all the 4 races in advance, you receive an error(pytest) that checks DB and sees orc, dwarf races (no players have those races in our json file, so they weren't created). But.. don't you think that it's smart to create all 4 races before player can chose from them? so our db after running a code MUST look like this: races: `elf, human` (because only those races players had in json). but I think it would be better like this: `elf, human, orc, dwarf.` (sadly tests fail in this case)
> 
> I hope you understood what I mean

Well the task and tests are designed so that you have to create races manually from file. I see your point but it would be a different task variation if to go by your logic.

main.py
panicua: > check return type here

yep, forgot that it can be None

fsocie7y (approved):
GJ 👍

PR #771, opened 2024-04-19
Solution

Chub-andrew:


YuriiKindrat (changes requested):
all tests should pass)

main.py
```diff
@@ -1,10 +1,52 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
```
fsocie7y, line 9: I'm not sure if it's an issue, but I would exit the `with` block as soon as I finished the last operation with the file. No need to keep useless references in memory, right?

fsocie7y (changes requested):

PR #772, opened 2024-04-20
Solution

y-kondrashova:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(unique=True)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```
YuriiKindrat, line 25: it common practice to use related name for FK

db/models.py
```diff
@@ -1 +1,26 @@
...
+    created_at = models.DateTimeField(auto_now=True)
```
YuriiKindrat, line 26: are u sure about auto_now here ?

main.py
```diff
@@ -1,10 +1,41 @@
...
-    pass
+    with open("players.json", "r") as players_file:
+        players = json.load(players_file)
+
+    for player, info in players.items():
+        guild_info = info["guild"]
+
+        race_info = info["race"]
+
+        guild = Guild.objects.get_or_create(
```
YuriiKindrat, line 17: u could write here like this:
`_, guild = Guild....` so there is won't be need to use indexes at 32 line

main.py
```diff
@@ -1,10 +1,44 @@
...
+
+        race_info = info["race"]
+
+        if guild_info:
+            guild, _ = Guild.objects.get_or_create(
+                name=guild_info["name"],
+                description=guild_info["description"],
+            )
+        else:
+            guild = None
```
fsocie7y, line 23: u can do it without else, move this line above)

fsocie7y (changes requested):

fsocie7y (approved):
GJ 👍

PR #778, opened 2024-04-22
solution

CharmNiovi:


db/models.py
```diff
@@ -1 +1,36 @@
...
+        Race,
+        on_delete=models.CASCADE,
+        related_name="players"
+    )
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        null=True,
+        default=None,
+        related_name="members"
```
fsocie7y, line 34: i think its better too keep related names the same)

fsocie7y (changes requested):

fsocie7y (approved):
GJ 👍

PR #790, opened 2024-04-24
solution

Ruxxell54:


db/migrations/0004_alter_skill_race.py
Oleksl888, line 1: Tests are failing because you didnt include all the migrations (they depend on each other as a chain)

db/models.py
```diff
@@ -1 +1,48 @@
 from django.db import models
+
+
+class Race(models.Model):
+    names = [
+        ("elf", "Elf"),
```
Oleksl888, line 6: not all races are included.Also it is a convention to call Choices options with capital letters and add _CHOICES to the end to distinguish from regular fields

main.py
```diff
@@ -1,10 +1,70 @@
 import init_django_orm  # noqa: F401
+import json
```
Oleksl888, line 2: check import order

main.py
```diff
@@ -1,10 +1,70 @@
...
+
+
+def create_or_update_guild(player_guild: dict) -> Guild | None:
+    if player_guild:
+        guild, _ = Guild.objects.get_or_create(
+            name=player_guild["name"],
+            defaults={"description": player_guild["description"]}
+        )
+        return guild
+    return None
```
Oleksl888, line 38: no need to return None explicitly - if you dont specify return, function will return None by default

Oleksl888 (changes requested):

PR #791, opened 2024-04-24
Solution

Ruxxell54:


db/models.py
```diff
@@ -1 +1,50 @@
 from django.db import models
+
+
+class Race(models.Model):
+    names_choices = [
```
Oleksl888, line 5: NAMES_CHOICES should be capitalized it is a convention to signify that this sequence is a constant and to differentiate from regular fields. Also first element in the tuple must match the names for races in players.json file

Oleksl888 (changes requested):

db/models.py
```diff
@@ -1 +1,49 @@
 from django.db import models
+
+
+class Race(models.Model):
+    NAMES_CHOICES = [
+        ("ELF", "Elf"),
```
Oleksl888, line 6: the first element in tuple here should match exactly the name that you would expect from your file ('elf', 'dwarf' etc) and that will be stored in db, second can be a human readable name which you will display to user

Oleksl888 (changes requested):

main.py
```diff
@@ -1,11 +1,65 @@
...
+
+    for player_name, player_data in data.items():
+        race_data = player_data["race"]
+        guild_data = player_data["guild"]
 
+        race = create_or_update_race(race_data)
+        create_or_update_skills(race_data["skills"], race)
+        guild = create_or_update_guild(guild_data)
 
-if __name__ == "__main__":
```
Oleksl888, line 10: there was no requirement to remove this

Oleksl888 (approved):
Approved

PR #795, opened 2024-04-25
Solution

CrazyHilly:


main.py
```diff
@@ -1,10 +1,43 @@
...
+        )
+
+        for skill in race["skills"]:
+            Skill.objects.get_or_create(
+                name=skill["name"],
+                bonus=skill["bonus"],
+                defaults={"race": race_obj},
+            )
+
+        guild_obj = Guild.objects.get_or_create(
```
vasylhnatiuk, line 29:
```suggestion
        guild_obj, _ = Guild.objects.get_or_create(
```

main.py
```diff
@@ -1,10 +1,43 @@
...
+            name=guild["name"],
+            description=guild["description"],
+        )[0] if guild else None
```
vasylhnatiuk, line 32:
```suggestion
        ) if guild else None
```

vasylhnatiuk (changes requested):

main.py
CrazyHilly: @vasylhnatiuk There is a piece of code after the brackets "if guild else None". If I replace "[0]" with "_", I will have to use something like "if guild else (None, None)" which looks worse in my opinion.

vasylhnatiuk: you use the same ` if guild else None`

vasylhnatiuk (changes requested):

CrazyHilly:
image

main.py
vsmutok, line 32:
```suggestion
        if guild is not None:
            guild_obj, _ = Guild.objects.get_or_create(
                name=player_info["guild"]["name"],
                description=player_info["guild"]["description"]
            )
        else:
            guild_obj = None
```

vsmutok (changes requested):

CrazyHilly:
Can you please clearly explain why I should make any changes to this piece of code? Because I cannot understand why using [0] is so bad. The code is compact, readable, understandable, and it's working. While first suggestion here doesn't work, and the second one is simply not pretty. 
I can write something like:
guild_obj = None
if guild:
      guild_obj, _ = Guild.objects.get_or_create(
            name=guild["name"],
            description=guild["description"],
        )
 But can someone please explain to me why it's better to replace [0] with this piece of code?

PR #797, opened 2024-04-26
Solution

dkibalenko:


db/migrations/0006_alter_skill_race.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

main.py
```diff
@@ -1,10 +1,41 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as json_file:
+        players_data = json.load(json_file)
+
+    for player_name, player_data in players_data.items():
+        race, is_race_created = Race.objects.get_or_create(
```
Dimosphen1, line 13: There is no need to create `is_race_created` variable in case it is not used. You can skip it here and in the other places as well
```suggestion
        race, _ = Race.objects.get_or_create(
```

db/models.py
```diff
@@ -1 +1,44 @@
...
+    def __str__(self) -> str:
+        return f"{self.name} - {self.bonus}"
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+    def __str__(self) -> str:
+        return f"Guild: {self.name} {self.description or ''}"
```
Dimosphen1, line 32: Do not include model name to the `__str__`, it is already included

db/models.py
```diff
@@ -1 +1,44 @@
 from django.db import models
+
+
+class Race(models.Model):
+    NAME_CHOICES = (
+        ("elf", "The elf race"),
+        ("dwarf", "The dwarf race"),
+        ("human", "Human race"),
+        ("ork", "The ork race"),
+    )
```
Dimosphen1, line 10: You could use `models.TextChoices` for better choices description

Dimosphen1 (changes requested):

main.py
```diff
@@ -1,10 +1,41 @@
...
-    pass
+    with open("players.json", "r") as json_file:
+        players_data = json.load(json_file)
+
+    for player_name, player_data in players_data.items():
+        race, _ = Race.objects.get_or_create(
+            name=player_data["race"]["name"],
+            description=player_data["race"]["description"]
+        )
+        if player_data["guild"]:
```
pavlopro, line 17: You will encounter an error if 'guild' is missing from the data

pavlopro (changes requested):

main.py
```diff
@@ -1,10 +1,68 @@
...
+        Skill.objects.get_or_create(name=name, bonus=bonus, race=race)
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as json_file:
+        players_data = json.load(json_file)
+
+    for player_name, player_info in players_data.items():
+        try:
```
pavlopro, line 59: Revert back as it was 
Why do you need try-except condition here
Just add a simple condition

```
guild = None
if player_data.get("guild"):
```

pavlopro (changes requested):

main.py
dkibalenko: I forgot about the possibility of a safe look-up with .get() in dict. I added input data validation and error handling logic as it's a proper general idea of building more robust functionality, isn't it? Although it might be redundant over here.

Polyakiv-Andrey (approved):
Good job!

PR #800, opened 2024-04-28
Implement Game models task

bwaah1:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
```
YuriiKindrat, line 26: what about related names ?
also why do u use Protect option here ?

YuriiKindrat (changes requested):

db/models.py
YuriiKindrat: also all tests should pass

bwaah1: > what about related names ? 
can you explain this in more detail?

bwaah1: > what about related names ? also why do u use Protect option here ?

in order to use related names, you need to change the test cases

main.py
```diff
@@ -1,10 +1,52 @@
 import init_django_orm  # noqa: F401
-
+import json
```
Oleksl888, line 2: this import must be above all since its build in library

main.py
```diff
@@ -1,10 +1,52 @@
...
+                skills_data = race_data["skills"]
+                for skill_data in skills_data:
+                    skill_name = skill_data["name"]
+                    skill_bonus = skill_data["bonus"]
+                    Skill.objects.get_or_create(
+                        name=skill_name,
+                        bonus=skill_bonus,
+                        race=race
+                    )
+            else:
```
Oleksl888, line 29: you can move race assignment above line 12 so you don't have to use extra `else` statement

main.py
```diff
@@ -1,10 +1,52 @@
...
+            guild_data = player_data["guild"]
+            if guild_data:
+                guild_name = guild_data["name"]
+                guild_description = guild_data["description"]
+                guild, _ = Guild.objects.get_or_create(
+                    name=guild_name,
+                    description=guild_description
+                )
+            else:
+                guild = None
```
Oleksl888, line 41: same here - refactor so you don't have to use `else`

PR #805, opened 2024-05-05
Solution

VasylRybalka:


main.py
```diff
@@ -1,10 +1,38 @@
...
+            guild_db = Guild.objects.get_or_create(
+                name=guild["name"], description=guild["description"])
+            guild_id = guild_db[0].id
+
+        if race:
+            race_db = Race.objects.get_or_create(
+                name=race["name"], description=race["description"])
+            race_id = race_db[0].id
+
+            if len(race["skills"]) > 0:
```
vasylhnatiuk, line 27:
```suggestion
            if race["skills"]:
```

main.py
```diff
@@ -1,10 +1,38 @@
...
+    for nickname, player in players.items():
+        guild = player["guild"]
+        race = player["race"]
+        guild_id = None
+        race_id = None
+
+        if guild:
+            guild_db = Guild.objects.get_or_create(
+                name=guild["name"], description=guild["description"])
+            guild_id = guild_db[0].id
```
vasylhnatiuk, line 20: it is not nessesary to get id, you can use object

vasylhnatiuk (changes requested):

PR #806, opened 2024-05-05
Solution

VladyslavKsh:


db/models.py
```diff
@@ -1 +1,28 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild,
+                              null=True,
+                              on_delete=models.SET_NULL)
```
YuriiKindrat, line 27: what about related names here ?

YuriiKindrat (changes requested):

db/models.py
```diff
@@ -1 +1,31 @@
...
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race,
+                             related_name="races",
+                             on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild,
+                              null=True,
+                              related_name="guilds",
+                              on_delete=models.SET_NULL)
```
fsocie7y, line 30: related names should be as class name in plural

fsocie7y (changes requested):

db/models.py
VladyslavKsh: Oh, yes...

fsocie7y (approved):
GJ 👍

PR #820, opened 2024-05-26
Solution

rinotokun:


main.py
```diff
@@ -1,10 +1,51 @@
...
+            name=skill["name"],
+            bonus=skill["bonus"],
+            race=race
+        )
+
+
+def create_guild(guild: dict) -> Guild | None:
+    if not guild:
+        return
+    guild_obj, created = Guild.objects.get_or_create(
```
Dimosphen1, line 27: You could omit using `created` variable by replacing it with `_`
```suggestion
    guild_obj, _ = Guild.objects.get_or_create(
```

main.py
```diff
@@ -1,10 +1,51 @@
 import init_django_orm  # noqa: F401
 
+import json
+
 from db.models import Race, Skill, Player, Guild
 
 
+def create_race(race: dict) -> Race | bool:
```
Dimosphen1, line 8: The return type annotation is incorrect, you indicate that you return with a `Race` class instance, or a boolean value. Actually, you return a tuple of these objects. Please, review your return type annotation or omit these methods and call these actions in the code directly

Dimosphen1 (changes requested):

vsmutok (approved):
Good Job!

PR #824, opened 2024-05-31
Solution

ShcherbynaSv:


db/models.py
```diff
@@ -1 +1,26 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
pavlopro, line 12: Add related name attribute for all relational fields

main.py
```diff
@@ -1,10 +1,42 @@
 import init_django_orm  # noqa: F401
```
pavlopro, line 1: Add blank line between different import types
Fix import type ordering

main.py
```diff
@@ -1,10 +1,42 @@
...
-    pass
+    Race.objects.all().delete()
+    Skill.objects.all().delete()
+    Player.objects.all().delete()
+    Guild.objects.all().delete()
+
+    with open("players.json", "r") as players_file:
+        players_data = json.load(players_file)
+
+    list_of_players = list(players_data)
```
pavlopro, line 16: There is no need to create this variable. Iterate over players_data directly.

main.py
```diff
@@ -1,10 +1,42 @@
...
+
+    for player in list_of_players:
```
pavlopro, line 18: It would be better to create `race` and `guid` separately, store them in variables, and then use them for `skill` and `player` creation.

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 5.0.6 on 2024-05-31 11:10
```

pavlopro (changes requested):
It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

PR #830, opened 2024-06-01
solved

serhii-chaban:


main.py
```diff
@@ -1,11 +1,39 @@
...
 
-if __name__ == "__main__":
-    main()
+        Player.objects.get_or_create(
+            nickname=player_name,
+            email=info.get("email"),
+            bio=info.get("bio"),
+            race=race_inst,
+            guild=guild_inst
+        )
```
vsmutok, line 39: here it is better to just use `create`, because you don't need to get the result

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍🏼

PR #837, opened 2024-06-07
solution

xtsvi:


main.py
```diff
@@ -1,10 +1,45 @@
...
+                race=race
+            )
+
+        Player.objects.get_or_create(
+            nickname=player_name,
+            email=player_info["email"],
+            bio=player_info["bio"],
+            race=race,
+            guild=guild
+        )
```
vsmutok, line 42: here it is better to use simply `create` if you do not need to save the result

main.py
```diff
@@ -1,10 +1,45 @@
...
+            guild = Guild.objects.get_or_create(
+                name=guild_dict["name"],
+                description=guild_dict["description"]
+            )[0]
+
+        race_dict = player_info["race"]
+        race = Race.objects.get_or_create(
+            name=race_dict["name"],
+            description=race_dict["description"]
+        )[0]
```
vsmutok, line 27: it looks a little better
```suggestion
        race, _ = Race.objects.get_or_create(
            name=race_dict["name"],
            description=race_dict["description"]
        )
```

vsmutok (changes requested):

vsmutok (approved):
Great 👍🏼

PR #840, opened 2024-06-11
implement game-models

daikj88:


main.py
```diff
@@ -1,10 +1,44 @@
...
+                race=race
+            )
+
+        Player.objects.get_or_create(
+            nickname=player_name,
+            email=player_data["email"],
+            bio=player_data["bio"],
+            race=race,
+            guild=guild
+        )
```
vsmutok, line 41: if you don't need to store the result of creation or get in a variable, then it's better to just use `create`

vsmutok (changes requested):

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2024-06-12 00:18
```

main.py
```diff
@@ -1,10 +1,44 @@
...
+            guild, _ = Guild.objects.get_or_create(
+                name=guild_data["name"],
+                description=guild_data["description"]
+            )
+        else:
+            guild = None
+
+        race_data = player_data.get("race")
+        race, _ = Race.objects.get_or_create(
+            name=race_data["name"],
```
pavlopro, line 24: Simplify this to directly retrieve `player_data["race"]["name"]` without declaring `race_data` variable.

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #844, opened 2024-06-15
Implement solution

anton-bilous:


db/models.py
```diff
@@ -1 +1,50 @@
...
+        return self.name
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+    )
```
MNDonut, line 18: don't forget to add related_name for ForeignKey and ManyToMany fields

main.py
```diff
@@ -1,10 +1,40 @@
...
 def main() -> None:
-    pass
+    with open("players.json") as f:
+        data = json.load(f)
+
+    for nickname, properties in data.items():
+        race_data = properties["race"]
+        skills = race_data.pop("skills")
+
+        race = Race.objects.get_or_create(**race_data)[0]
```
MNDonut, line 16: it's better to unpack result in two variables instead of using index

main.py
```diff
@@ -1,10 +1,40 @@
...
+
+        for skill_data in skills:
+            Skill.objects.get_or_create(
+                **skill_data,
+                race=race,
+            )
+
+        guild_data = properties["guild"]
+        guild = (
+            Guild.objects.get_or_create(**guild_data)[0]
```
MNDonut, line 26: unpack instead of index

MNDonut (changes requested):

db/models.py
anton-bilous: Why do I need to set it if the tests expect it to be default?

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #845, opened 2024-06-16
Solution

VladimirDolhyi:


main.py
```diff
@@ -1,10 +1,33 @@
...
+        guild_data = player_data["guild"]
+        guild = None
+        if guild_data:
+            guild, created = Guild.objects.get_or_create(
+                name=guild_data["name"], description=guild_data["description"]
+            )
+        Player.objects.get_or_create(
+            nickname=player, email=player_data["email"],
+            bio=player_data["bio"], race=race, guild=guild
+        )
```
vsmutok, line 30: here it is better to use simply `create` if you do not need to save the result

vsmutok (changes requested):

main.py
```diff
@@ -1,10 +1,33 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_file:
+        players = json.load(players_file)
+
+    for player, player_data in players.items():
+        race_data = player_data["race"]
+        race, created = Race.objects.get_or_create(
```
pavlopro, line 13: `created`
You will not use this variable. Replace it with `_`

main.py
```diff
@@ -1,10 +1,33 @@
...
+        )
+        for skill_data in race_data["skills"]:
+            Skill.objects.get_or_create(
+                name=skill_data["name"], bonus=skill_data["bonus"],
+                race=race
+            )
+        guild_data = player_data["guild"]
+        guild = None
+        if guild_data:
+            guild, created = Guild.objects.get_or_create(
```
pavlopro, line 24: The same here

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #851, opened 2024-06-20
sol

Unlie9:


main.py
```diff
@@ -1,11 +1,36 @@
...
+        get_rase, _ = Race.objects.get_or_create(
+            name=race_info["name"],
+            description=race_info["description"])
+
+        if guild_info:
+            get_guild, _ = Guild.objects.get_or_create(
+                name=guild_info["name"],
+                description=guild_info["description"])
+        else:
+            get_guild = None
```
DevSpaciX, line 23: You can create this viriable at the start and do not use else block .

main.py
```diff
@@ -1,11 +1,36 @@
...
-    pass
+    with open("players.json") as file:
+        players_info = json.load(file)
+
+    for nick, info in players_info.items():
+        race_info = info["race"]
+        guild_info = info["guild"]
+
+        get_rase, _ = Race.objects.get_or_create(
+            name=race_info["name"],
```
DevSpaciX, line 15: `get_rase` looks like func name , maybe call this viriable `race_data`  or smth like that . And yeah raSe ? Did you mean raCe ?)

main.py
```diff
@@ -1,11 +1,36 @@
...
+            description=race_info["description"])
+
+        if guild_info:
+            get_guild, _ = Guild.objects.get_or_create(
+                name=guild_info["name"],
+                description=guild_info["description"])
```
DevSpaciX, line 21: Same here with `get_guild` rename it pls

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #856, opened 2024-06-22
Solution v1.0

VShveda:


main.py
```diff
@@ -1,10 +1,48 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    Guild.objects.all().delete()
+    Skill.objects.all().delete()
+    Race.objects.all().delete()
+    Player.objects.all().delete()
```
DevSpaciX, line 11: What is this ?))

db/models.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=True)
```
DevSpaciX, line 22: Why did you make email field unique ?) I do not see that is necessery here , only for nickname 
image

DevSpaciX (changes requested):

main.py
```diff
@@ -1,10 +1,43 @@
...
+
+        for name, player_info in players.items():
+            race_info = player_info["race"]
+            race, _ = Race.objects.get_or_create(
+                name=race_info["name"],
+                description=race_info["description"]
+            )
+
+            skill_info = race_info["skills"]
+            for skill in skill_info:
```
DevSpaciX, line 19: Could you iterate without creating another viriable ?
`for skill in race_info['skills']:`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #858, opened 2024-06-23
solution

NikitaNikolaitsev:


main.py
```diff
@@ -1,10 +1,44 @@
...
+import json
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    # Skill.objects.all().delete()
+    # Player.objects.all().delete()
+    # Guild.objects.all().delete()
+    # Race.objects.all().delete()
```
DevSpaciX, line 10: Remove this

main.py
```diff
@@ -1,10 +1,44 @@
...
+                name=skill["name"],
+                bonus=skill["bonus"],
+                race=race,
+            )
+        if guild_info:
+            guild, _ = Guild.objects.get_or_create(
+                name=guild_info["name"],
+                description=guild_info["description"])
+        else:
+            guild = None
```
DevSpaciX, line 33: Ypu can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

DevSpaciX (changes requested):
Oh , and push migrations to the PR for fix tests ! Always check if tests are passed

DevSpaciX (approved):
Mega nice now !

PR #860, opened 2024-06-23
Solution

b3v3kt0r:


db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255)
+    email = models.EmailField(max_length=255)
```
DevSpaciX, line 22: Why you did not make nickname unique ?) Pay more attention to all detaild of the task 
image

main.py
```diff
@@ -1,10 +1,41 @@
...
+                                        bonus=skill["bonus"],
+                                        race=race)
+
+        if guild_data:
+            guild, _ = (
+                Guild.objects.get_or_create(
+                    name=guild_data["name"],
+                    description=guild_data["description"]))
+        else:
+            guild = None
```
DevSpaciX, line 32: You can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #862, opened 2024-06-23
Solution

Yuruch:


db/models.py
```diff
@@ -1 +1,31 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=True)
```
DevSpaciX, line 22: Why did you make `email` unique ? Is the nickname only was required for being unique , is not it ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #863, opened 2024-06-23
Solution

SrMusienko:


main.py
```diff
@@ -1,10 +1,37 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+    for user_name in data:
+        user = data[user_name]
+        race, flag = Race.objects.get_or_create(
+            name=user["race"]["name"],
+            description=user["race"]["description"])
```
DevSpaciX, line 13: If you do not using `flag` viriable , just unpuck it into a `_ `

main.py
```diff
@@ -1,10 +1,37 @@
...
+
+        for skill_data in user["race"]["skills"]:
+            Skill.objects.get_or_create(
+                name=skill_data["name"],
+                bonus=skill_data["bonus"], race=race
+            )
+
+        if user.get("guild"):
+            guild, flag = Guild.objects.get_or_create(
+                name=user["guild"]["name"],
```
DevSpaciX, line 23: same here with flag : )

main.py
```diff
@@ -1,10 +1,37 @@
...
+                description=user["guild"]["description"])
+        else:
+            guild = None
```
DevSpaciX, line 26: You can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #865, opened 2024-06-24
Solution V.1

SemenetsSergii:


main.py
```diff
@@ -1,10 +1,43 @@
...
+                )
+
+            guild_info = person["guild"]
+            if guild_info:
+                guild, created = Guild.objects.get_or_create(
+                    name=guild_info["name"],
+                    description=guild_info["description"]
+                )
+            else:
+                guild = None
```
DevSpaciX, line 32: You can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #866, opened 2024-06-24
Solution

AntonKorinchuk:


main.py
```diff
@@ -1,10 +1,41 @@
...
+                race=race
+            )
+
+        if player_info["guild"]:
+            guild, _ = Guild.objects.get_or_create(
+                name=player_info["guild"]["name"],
+                description=player_info["guild"]["description"]
+            )
+        else:
+            guild = None
```
DevSpaciX, line 30: You can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #867, opened 2024-06-24
Sol

Oleh323:


main.py
```diff
@@ -1,10 +1,38 @@
...
+                name=skill_data["name"],
+                bonus=skill_data["bonus"], race=race
+            )
+
+        if user.get("guild"):
+            guild, flag = Guild.objects.get_or_create(
+                name=user["guild"]["name"],
+                description=user["guild"]["description"])
+        else:
+            guild = None
```
DevSpaciX, line 27: You can create this viriable at the start and do not use else block .

main.py
```diff
@@ -1,10 +1,38 @@
...
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        data = json.load(file)
+    for user_name in data:
+        user = data[user_name]
+        race, flag = Race.objects.get_or_create(
+            name=user["race"]["name"],
+            description=user["race"]["description"])
```
DevSpaciX, line 14: If you do not using `flag` viriable , just unpuck it into a `_ `

main.py
```diff
@@ -1,10 +1,38 @@
...
+
+        for skill_data in user["race"]["skills"]:
+            Skill.objects.get_or_create(
+                name=skill_data["name"],
+                bonus=skill_data["bonus"], race=race
+            )
+
+        if user.get("guild"):
+            guild, flag = Guild.objects.get_or_create(
```
DevSpaciX, line 23: same here with flag : )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #869, opened 2024-06-24
test push in git

TkachukTara:


main.py
```diff
@@ -1,10 +1,46 @@
...
+        race = player["race"]
+        guild_id = None
+        race_id = None
+
+        if guild:
+            guild_db = Guild.objects.get_or_create(
+                name=guild["name"],
+                description=guild["description"]
+            )
+            guild_id = guild_db[0].id
```
DevSpaciX, line 23: You can unpuck your rusult into a two viriables , one of them will be instance and second is `created` with can be `_ `, cuz we do not use it . That way it will be more clean imho

main.py
```diff
@@ -1,10 +1,46 @@
...
+                    Skill.objects.get_or_create(
+                        name=skill["name"],
+                        bonus=skill["bonus"],
+                        race_id=race_id)
+
+        Player.objects.get_or_create(nickname=nickname,
+                                     email=player["email"],
+                                     bio=player["bio"],
+                                     race_id=race_id,
+                                     guild_id=guild_id)
```
DevSpaciX, line 43: And here you can try to paste your instances instead of ids witch you got by indexes

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #871, opened 2024-06-24
sol1

IvanVahanov:


main.py
```diff
@@ -1,10 +1,48 @@
...
+                }
+            )
+
+        if guild_data:
+            guild, created = Guild.objects.get_or_create(
+                name=guild_data["name"],
+                defaults={"description": guild_data["description"]}
+            )
+        else:
+            guild = None
```
DevSpaciX, line 35: You can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

PR #872, opened 2024-06-24
'Solution'

Irina17191:


db/models.py
```diff
@@ -1 +1,38 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255, unique=True)
```
DevSpaciX, line 14: Why did you make bonus unique ?
In task only name was required

main.py
```diff
@@ -1,10 +1,58 @@
...
+        guild_data = player_data.get("guild")
+        if guild_data:
+            guild_name = guild_data.get("name", "")
+            guild_description = guild_data.get("description", None)
+            guild, _ = Guild.objects.get_or_create(
+                name=guild_name,
+                defaults={"description": guild_description}
+            )
+        else:
+            guild = None
```
DevSpaciX, line 30: You can create this viriable at the start and do not use else block .

DevSpaciX (changes requested):

main.py
```diff
@@ -1,10 +1,58 @@
...
+        player, _ = Player.objects.get_or_create(
+            nickname=nickname,
+            email=email,
+            bio=bio,
+            race=race,
+            guild=guild
+        )
+
+        print(f"Created player: {player.nickname}")
+
```
DevSpaciX, line 51: Remove this print too : )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #883, opened 2024-07-04
Solution

IraMladanovych02:


main.py
```diff
@@ -1,10 +1,54 @@
...
+    with open("players.json", "r") as file:
+        players_data = json.load(file)
+
+    for player_name, player_data in players_data.items():
+        print(player_data)
+
+        race_name = player_data["race"]["name"]
+        race_description = player_data["race"].get("description", " ")
+
+        race, created = Race.objects.get_or_create(
```
Dimosphen1, line 17: Since you do not use `created`, you could omit it here and in the other places
```suggestion
        race, _ = Race.objects.get_or_create(
```

main.py
```diff
@@ -1,10 +1,54 @@
...
+            name=race_name,
+            defaults={"description": race_description}
+        )
+
+        for skill_data in player_data["race"].get("skills", []):
+            skill_name = skill_data["name"]
+            skill_bonus = skill_data["bonus"]
```
Dimosphen1, line 24: There is no need to create these variables, pass data directly to `get_or_create` here and in the other places

Dimosphen1 (changes requested):
Several changes were requested. Make sure that all tests pass (you may need to include migration for that)

main.py
```diff
@@ -1,10 +1,50 @@
...
+        guild = None
+        if guild_data:
+            guild_name = guild_data["name"]
+            guild_description = guild_data.get("description")
+            guild, created = Guild.objects.get_or_create(
+                name=guild_name,
+                defaults={"description": guild_description}
+            )
+
+        Player.objects.get_or_create(
```
vsmutok, line 38: if you don't need to check the object and whether it was created, then it's better to just use "create" in this case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #886, opened 2024-07-06
implement py-game-models

HalynaRohatska:


main.py
```diff
@@ -2,9 +2,49 @@
 
 from db.models import Race, Skill, Player, Guild
 
+import json
```
oliasymak24, line 5: Change the order of your imports according to `pep8`. Here `json` should be first.

main.py
```diff
@@ -2,9 +2,49 @@
...
+                description=data_guild["description"]
+            )
+
+        Player.objects.get_or_create(
+            nickname=key,
+            email=value["email"],
+            bio=value["bio"],
+            race=get_race,
+            guild=get_guild
+        )
```
oliasymak24, line 47: For better readability suggest split to smaller functions

main.py
```diff
@@ -2,9 +2,49 @@
...
+        data = json.load(file)
+    return data
+
 
 def main() -> None:
-    pass
+    path_to_file = "players.json"
+    data_players = load_players(path_to_file)
+
+    for key, value in data_players.items():
```
oliasymak24, line 18: Better to use more descreptive names of variable, check what it is key and value in yout json file and change it)

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #889, opened 2024-07-10
+

AndriZhok:


main.py
```diff
@@ -1,10 +1,41 @@
...
+                )
+
+            guild_info = player_info["guild"]
+            guild = None
+            if guild_info:
+                guild, _ = Guild.objects.get_or_create(
+                    name=guild_info["name"],
+                    description=guild_info["description"]
+                )
+            Player.objects.get_or_create(
```
vsmutok, line 32: it's better to just use `create` here if you don't need to get the result

vsmutok (changes requested):

main.py
```diff
@@ -1,10 +1,41 @@
 import init_django_orm  # noqa: F401
-
+import json
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        players = json.load(file)
```
Dimosphen1, line 8: Do not overload a context manager as mentioned in the task description

db/models.py
```diff
@@ -1 +1,30 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Dimosphen1, line 24: Add `related_name` to FK (`ForeignKey`) fields

Dimosphen1 (changes requested):

db/models.py
```diff
@@ -1 +1,34 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="rase"
```
Dimosphen1, line 27: Related name is not correct, you should define how you would retrieve players from `Race` instance. `players` would be an appropriate name here and for `guild` field as well

db/migrations/0002_alter_player_race.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

Dimosphen1 (changes requested):

main.py
```diff
@@ -1,10 +1,41 @@
...
+            )
+
+        guild_info = player_info["guild"]
+        guild = None
+        if guild_info:
+            guild, _ = Guild.objects.get_or_create(
+                name=guild_info["name"],
+                description=guild_info["description"]
+            )
+        Player.objects.get_or_create(
```
pavlopro, line 32: it's better to just use create here if you don't need to get the result

pavlopro (changes requested):

pavlopro (approved):
Good Job!

pavlopro (approved):
Good Job!

PR #892, opened 2024-07-14
Develop

TsvetanKichuk:


pavlopro (changes requested):
Tests Failed

.flake8
```diff
@@ -6,5 +6,5 @@ max-complexity = 18
 select = B,C,E,F,W,T4,B9,ANN,Q0,N8,VNE
 exclude =
     **migrations
-    venv
+    venv .venv
```
pavlopro, line 9: Do not modify `.flake8` file

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2024-07-11 20:40
```

main.py
```diff
@@ -1,10 +1,51 @@
...
-    pass
+    with open("players.json", "r") as f:
+        player_data = json.load(f)
+
+    for player_name, player_info in player_data.items():
+        email = player_info["email"]
+        bio = player_info["bio"]
+
+        race_data = player_info["race"]
+        race, created_race = Race.objects.get_or_create(
```
pavlopro, line 17: You don't use `created_race` variable. Can be simplified:
`race, _ =` ...

main.py
```diff
@@ -1,10 +1,51 @@
...
+            name=race_data["name"],
+            description=race_data["description"]
+        )
+        print(Race.objects.all())
```
pavlopro, line 21: Remove print in all places

main.py
```diff
@@ -1,10 +1,51 @@
...
+
+        skills = []
+        for skill_data in race_data.get("skills", []):
+            skill, created_skill = Skill.objects.get_or_create(
+                name=skill_data["name"], race=race, bonus=skill_data["bonus"]
+            )
+            skills.append(skill)
+        print(Skill.objects.all())
```
pavlopro, line 29: Simplify Skills creation
There is no need to store skill into variable.
There is no need to append it to list

main.py
```diff
@@ -1,10 +1,51 @@
...
+
+        guild_data = player_info.get("guild")
+        guild = None
+        if guild_data:
+            guild, created_guild = Guild.objects.get_or_create(
```
pavlopro, line 34: Simplify to:
`guild, _ =` ...

pavlopro (changes requested):

db/migrations/0001_initial.py
pavlopro: Leave only one migration file

pavlopro (changes requested):

pavlopro (changes requested):
Please remove db file from PR

pavlopro (approved):
Good Job!

PR #895, opened 2024-07-15
game models solution

TenbReala:


pavlopro (changes requested):
Tests Failed

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2024-07-14 19:55
```

pavlopro (changes requested):

vsmutok (approved):
Well Done!

PR #901, opened 2024-07-21
Solution

Wallian169:


main.py
```diff
@@ -1,10 +1,61 @@
...
+        return None
+    guild, created = Guild.objects.get_or_create(
+        name=guild_data["name"],
+        defaults={"description": guild_data.get("description", "")}
+    )
+    return guild
+
+
+def create_prayers(data: dict) -> None:
+    for name, details in data.items():
```
pavlopro, line 40: Good, but please simplify this and make all actions right in `create_prayers` function. There no need to have those additional functions.

main.py
```diff
@@ -1,10 +1,61 @@
...
+                "email": details["email"],
+                "bio": details["bio"],
+                "race": race,
+                "guild": guild
+            }
+        )
+
+
 def main() -> None:
-    pass
```
pavlopro, line 7: Do not made any changes in `main` function

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2024-07-21 11:49
```

pavlopro (changes requested):

PR #905, opened 2024-07-22
Develop

b4oody:


db/migrations/0003_alter_player_guild.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR in the future, please

main.py
```diff
@@ -1,10 +1,41 @@
...
+
+    for player, values in players_json_file.items():
+        race, _ = Race.objects.get_or_create(
+            name=values["race"]["name"],
+            description=values["race"]["description"]
+        )
+
+        guild = None
+        guild_info = values.get("guild")
+        if guild_info:
```
Dimosphen1, line 20: You could write this condition in one line using walrus operator
```suggestion
        if guild_info := values.get("guild"):
```

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Dimosphen1, line 24: Add `related_name` attribute to the ForegnKey or Many-to-Many fields in the future

Dimosphen1 (approved):
LGTM, several comments were left for consideration

PR #915, opened 2024-08-02
Update

Vsevoloddeez:


main.py
```diff
@@ -1,10 +1,40 @@
...
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players:
+        players = json.load(players)
+    for key in players:
```
vasylhnatiuk, line 11: key is a bad name

vasylhnatiuk (changes requested):

main.py
Vsevoloddeez: aaan what do you want me to do ?

EdAlekseiev (approved):

PR #925, opened 2024-08-11
Solution

OleksiukStepan:


settings.py
oliasymak24, line 1: Don't need to commit this

main.py
```diff
@@ -1,11 +1,51 @@
...
-    main()
+    Player.objects.get_or_create(
+        nickname=player_nickname,
+        race=race,
+        guild=guild,
+        defaults={
+            "email": player_info.get("email"),
+            "bio": player_info.get("bio"),
+        }
+    )
```
oliasymak24, line 51: Must be in main fucntion too

oliasymak24 (changes requested):

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    create_at = models.DateTimeField(auto_now=True)
```
MNDonut, line 26: are you sure that **auto_now** should be used here?
Also field name is incorrect

MNDonut (changes requested):

oliasymak24 (approved):

PR #928, opened 2024-08-12
completed

BGBroth:


db/migrations/0002_alter_player_guild.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

main.py
```diff
@@ -1,10 +1,46 @@
...
+            for skill_data in race_data.get("skills"):
+                Skill.objects.create(
+                    name=skill_data["name"],
+                    bonus=skill_data["bonus"],
+                    race=race
+                )
+
+        guild = None
+        if guild_data:
+            guild, created = Guild.objects.get_or_create(
```
Dimosphen1, line 30: Since you do not use `created` variable here and on line 35, you could omit it with `_`

Dimosphen1 (changes requested):
Several minor changes were requested

PR #930, opened 2024-08-16
Solution

dartomOOv:


main.py
```diff
@@ -1,10 +1,44 @@
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
+import json
```
oliasymak24, line 4: Change the order of your imports according to `pep8`. Here `json` should be first.

oliasymak24 (approved):
Good job!
p.s. some small comment)

PR #931, opened 2024-08-16
Solution

comnotaDidi:


main.py
```diff
@@ -1,10 +1,43 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_info:
```
pavlopro, line 9: Do not overload the context manager.
The context manager is needed to work with the file, in our case, to read data. Therefore, after you have read the data from the file, exit the block. Read the checklist

main.py
```diff
@@ -1,10 +1,43 @@
...
+            person_bio = value["bio"]
+            if value["guild"] is not None:
+                Guild.objects.get_or_create(
+                    name=value["guild"]["name"],
+                    description=value["guild"]["description"],
+                )
+                guild_name = Guild.objects.get(name=value["guild"]["name"])
+            else:
+                guild_name = None
+            Race.objects.get_or_create(
```
pavlopro, line 22: Store the result into variable and use instead `race_name`

main.py
```diff
@@ -1,10 +1,43 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_info:
+        players = json.load(players_info)
+        for key, value in players.items():
+            person_email = value["email"]
+            person_bio = value["bio"]
```
pavlopro, line 13: guild = None

main.py
```diff
@@ -1,10 +1,43 @@
...
+            if value["guild"] is not None:
+                Guild.objects.get_or_create(
```
pavlopro, line 15: Store it into variable and use isntead of `guild_name`

pavlopro (changes requested):
Do not delete __init__.py files

db/__init__.py
Dimosphen1, line 1: Do not remove `__init__` files, make sure that they are not included in the PR (pull request)

main.py
```diff
@@ -1,10 +1,45 @@
...
+                description=player_data["race"]["description"],
+            )
+            for skill in player_data["race"]["skills"]:
+                race_name = Race.objects.get(name=player_data["race"]["name"])
+                Skill.objects.get_or_create(
+                    name=skill["name"],
+                    bonus=skill["bonus"],
+                    race=race_name
+                )
+            race_name = Race.objects.get(name=player_data["race"]["name"])
```
Dimosphen1, line 35: There is no need to retrieve Race one more time, you could save the value returned on line 24

main.py
Dimosphen1: Do not overload a context manager, refer to a checklist

main.py
Dimosphen1: Do not retrieve data from `Guild` twice, do it once, avoid operation on line 19.

Dimosphen1 (changes requested):

PR #935, opened 2024-08-18
Solution

DSahalatyi:


db/models.py
```diff
@@ -1 +1,26 @@
...
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(unique=True, max_length=255)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(unique=True, max_length=255)
```
MNDonut, line 21: there is a good rule that we should pass the required params first. in your case max_length is a required param while unique isn't

MNDonut (changes requested):

main.py
```diff
@@ -1,10 +1,64 @@
...
+                bonus=skill["bonus"],
+                race=player_race
+            )
+
+        player_guild = None
+        if data["guild"] is not None:
+            player_guild = Guild.objects.get_or_create(
+                name=data["guild"]["name"],
+                description=data["guild"]["description"]
+            )[0]
```
MNDonut, line 53: instead of indexes let's unpack the results of get_or_create method(it's a good practice)

MNDonut (changes requested):

MNDonut (approved):

PR #937, opened 2024-08-18
solution

BastovOleksandr:


db/models.py
```diff
@@ -1 +1,26 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255, unique=True)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE, null=True)
```
oliasymak24, line 12: I'm not sure that can be null

main.py
```diff
@@ -1,10 +1,39 @@
...
+            nickname=player_name,
+            email=player_info["email"],
+            bio=player_info["bio"],
+            race=race,
+
+            guild=Guild.objects.get_or_create(
+                name=player_info["guild"]["name"],
+                description=player_info["guild"]["description"]
+            )[0] if player_info["guild"] else None
+        )
```
oliasymak24, line 28: Separate this code

main.py
oliasymak24, line 1: Split into smaller functions and call it in main

main.py
```diff
@@ -1,10 +1,39 @@
...
-    pass
+
+    with open("players.json", "r") as file:
+        players = json.load(file)
+
+    for player_name, player_info in players.items():
+        race = Race.objects.get_or_create(
+            name=player_info["race"]["name"],
+            description=player_info["race"]["description"]
+        )[0]
```
oliasymak24, line 16: instead of using indexes, can just put `race, _`, check in other places

oliasymak24 (approved):

PR #945, opened 2024-08-25
Solution Politov

Alexandr-Politov:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
```
oliasymak24, line 6: Can be blank, not null

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255, blank=True, null=True)
```
oliasymak24, line 23: Where you see that can be blank and null?

main.py
```diff
@@ -1,11 +1,51 @@
 import init_django_orm  # noqa: F401
+import json
```
oliasymak24, line 2: Change the order of your imports according to `pep8`. Here `json` should be first.

main.py
```diff
@@ -1,11 +1,51 @@
...
+        guild = Guild.objects.get(
+            name=players_info[player]["guild"]["name"]
+        ) if players_info[player]["guild"] else None
+        Player.objects.create(
+            nickname=player,
+            email=players_info[player]["email"],
+            bio=players_info[player]["bio"],
+            race=race,
+            guild=guild
+        )
```
oliasymak24, line 51: All these stuff also must be in main
I suggest split it into small functions and call everything in main

main.py
```diff
@@ -1,11 +1,51 @@
...
+        players_info = json.load(file)
 
+    races_list = []
+    for user in players_info:
+        if players_info[user]["race"] not in races_list:
+            races_list.append(players_info[user]["race"])
 
-if __name__ == "__main__":
-    main()
+    for race in races_list:
```
oliasymak24, line 16: How can you access to races_list if you create it in main function?

main.py
oliasymak24, line 51: use the get_or_create() methodm check in other places too

oliasymak24 (changes requested):

db/models.py
MNDonut: still not fixed

main.py
```diff
@@ -1,11 +1,72 @@
+import json
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
+def create_races() -> None:
+    with open("players.json", "r") as file:
+        players_info = json.load(file)
```
MNDonut, line 9: let's open the file in the main function to not do it every time because it may be a large file with 10 gigabytes of data

MNDonut (changes requested):

main.py
```diff
@@ -1,11 +1,65 @@
...
+            defaults={
+                "email": players_info[player]["email"],
+                "bio": players_info[player]["bio"],
+                "race": race,
+                "guild": guild
+            }
+        )
+
+
 def main() -> None:
```
oliasymak24, line 57: Where you call main function?

oliasymak24 (approved):

PR #947, opened 2024-08-25
Solution

mihavryliuk:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(unique=True, max_length=255,)
```
MNDonut, line 5: it's better to pass max_length as the first param always because it's a required param. also the last comma is redundant

MNDonut (changes requested):

oliasymak24 (approved):

PR #948, opened 2024-08-25
Task Solution

achalaidiuk:


db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(unique=True)
```
MNDonut, line 22: are you sure you need unique param here?

MNDonut: and what about max_length?

db/models.py
```diff
@@ -1 +1,26 @@
...
+    bio = models.TextField(blank=True)
+    race = models.ForeignKey(Race, on_delete=models.PROTECT)
```
MNDonut, line 24: it's better to use CASCADE because PROTECT has some disadvantages like you won't be able to delete race(for ex. it's canceled) because some players exist

db/models.py
```diff
@@ -1 +1,26 @@
...
+    guild = models.ForeignKey(Guild, null=True, on_delete=models.SET_NULL)
```
MNDonut, line 25: for FK we firstly pass to= and on_delete= and then other params

MNDonut (changes requested):

db/models.py
```diff
@@ -19,8 +19,8 @@ class Guild(models.Model):
 
 class Player(models.Model):
     nickname = models.CharField(max_length=255, unique=True)
-    email = models.EmailField(unique=True)
+    email = models.EmailField(max_length=255)
     bio = models.TextField(blank=True)
-    race = models.ForeignKey(Race, on_delete=models.PROTECT)
-    guild = models.ForeignKey(Guild, null=True, on_delete=models.SET_NULL)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```
MNDonut, line 25: Did you push migrations? :)

MNDonut (changes requested):

db/models.py
achalaidiuk: It slipped my mind, I migrated).

MNDonut (approved):

PR #949, opened 2024-08-25
first_try

yashkunn:


db/models.py
```diff
@@ -1 +1,37 @@
...
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE
+    )
```
MNDonut, line 15: 1. doesn't it fit in 1 line
2. I see you add related name everywhere, so let's add it here as well

MNDonut (changes requested):

MNDonut (approved):

PR #950, opened 2024-08-25
Solution

panichqa:


db/models.py
```diff
@@ -1 +1,35 @@
...
+    description = models.TextField(blank=True)
+
+    def __str__(self) -> str:
+        return self.name
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE, null=True)
```
MNDonut, line 15: race cannot be null because otherwise it means skill can be created without the race that is wrong + it wasn't mentioned in the requirements

main.py
```diff
@@ -1,10 +1,49 @@
...
+        players_data = json.load(f)
+
+    for player in players_data:
+
+        race_description = players_data[player]["race"]["description"]
+        race_obj, _ = Race.objects.get_or_create(
+            name=players_data[player]["race"]["name"],
+            description=race_description
+            if race_description is not None
+            else ""
```
MNDonut, line 18: you can simplify this statement using "or" operator

MNDonut (changes requested):

main.py
```diff
@@ -1,10 +1,48 @@
...
+
+        if players_data[player]["guild"]:
+            guild_name = players_data[player]["guild"]["name"]
+            guild_description = players_data[player]["guild"]["description"]
+            guild_obj, _ = Guild.objects.get_or_create(
+                name=guild_name,
+                description=guild_description
+            )
+        else:
+            guild_obj = None
```
oliasymak24, line 35: You don't need this else here, just put guid_obj before your `if`

main.py
```diff
@@ -1,10 +1,48 @@
...
 from db.models import Race, Skill, Player, Guild
+import json
 
 
 def main() -> None:
     pass
+    with open("players.json") as f:
+        players_data = json.load(f)
+
+    for player in players_data:
```
oliasymak24, line 12: use items()

oliasymak24 (changes requested):

MNDonut (approved):

PR #951, opened 2024-08-26
Solution

wataschi:


db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2024-08-26 14:30
```

db/models.py
```diff
@@ -1 +1,28 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE,
```
pavlopro, line 12: No need to have those related name as 'skill_set' because by default django use class name + "_set"

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #952, opened 2024-08-26
'Solution'

ABaidin:


main.py
oliasymak24, line 1: suggest split into small functions and call them in main

oliasymak24 (approved):

PR #953, opened 2024-08-26
Develop

mnezbrytska:


main.py
```diff
@@ -1,10 +1,51 @@
...
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players_data = json.load(file)
+        print(players_data)
```
oliasymak24, line 11: Delete this print

main.py
```diff
@@ -1,10 +1,51 @@
...
+                "email": player_data["email"],
+                "bio": player_data["bio"],
+                "race": race,
+                "guild": guild,
+            }
+        )
+
+        if created:
+            player.skills.set(skills)
+            player.save()
```
oliasymak24, line 48: You don't need this part

main.py
```diff
@@ -1,10 +1,51 @@
...
+            defaults={
+                "email": player_data["email"],
+                "bio": player_data["bio"],
+                "race": race,
+                "guild": guild,
+            }
+        )
+
+        if created:
+            player.skills.set(skills)
```
oliasymak24, line 47: Does player need skills in the task?

oliasymak24 (changes requested):

db/models.py
```diff
@@ -1 +1,40 @@
...
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="skill_set"
```
MNDonut, line 15: you set a default related_name, so it doesn't make any sense.
If you didn't add related_name parameter, then you can access skills of any race using
race.skill_set.all(). However you set the same related_name, this is why it doesn't make any sense.
It'd be better to set something like "skills", then you can access them using:
race.skills.all(). This is the purpose of related_name parameter

db/models.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="race_players"
```
MNDonut, line 31: Not a mistake, but it's better to make this related_name like "players"
In this case when you access some Race object, you can get its players using:
race.players.all() instead of race.race_players.all(). Now you can see that there is no need to add "race_"
The same for guild attribute

db/models.py
```diff
@@ -1 +1,40 @@
...
+    )
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        related_name="guild_players",
+        null=True, blank=True
+    )
+    skills = models.ManyToManyField(Skill, related_name="players")
```
MNDonut, line 39: player shouldn't have skills attribute. check the requirements again

main.py
```diff
@@ -1,10 +1,46 @@
...
+        players_data = json.load(file)
+
+    for nickname, player_data in players_data.items():
+        race_data = player_data["race"]
+        race, _ = Race.objects.get_or_create(
+            name=race_data["name"],
+            defaults={"description": race_data["description"]}
+        )
+
+        skills = []
```
MNDonut, line 19: where do you use this variable?

MNDonut (changes requested):

db/models.py
mnezbrytska: Yes, you're right, in this case I use "set" as "box" and didn't think about the default name. (but using set is not so easy, I tried it in another task and it didn't work, only with related name worked well)

mnezbrytska: Yes you're right, in this case I use "set" as "box" and didn't think about the default name. i can't change this related name becouse in tests use skill_set. 
(but using set is not so easy, I tried it in another task and it didn't work, only with related name worked well, it's kind of magic)

db/models.py
```diff
@@ -1 +1,38 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE
```
Oleksl888, line 14: Are you sure you have to delete a skill if some Race gets deleted? What if it's used by other Race?

    This is what checklist says:
   

>  2) Use everything for its intended purpose
> 
>         Don't confuse the functionality of the on_delete parameter
>         Use on_delete=models.CASCADE if you know that some model doesn't exist without other model. So when the referenced object is deleted, also delete the objects that have references to it.
> 
>     Use on_delete=models.SET_NULL if you know that the object should not be deleted, even if the object it references may be deleted.
> 
>

db/models.py
```diff
@@ -1 +1,38 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```
Oleksl888, line 6: I guess the field can be null in DB as well so it's all aligned

db/models.py
```diff
@@ -1 +1,38 @@
...
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE
+    )
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
```
Oleksl888, line 20: Would make sense to make it blank also so ORM is aligned with DB

db/models.py
```diff
@@ -1 +1,38 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="players_of_race"
```
Oleksl888, line 30: `players` would be a more appropriate name - this is how a players will be referenced from race model

db/models.py
```diff
@@ -1 +1,38 @@
...
+    )
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        related_name="players_of_guild",
```
Oleksl888, line 35: same as above

main.py
```diff
@@ -1,10 +1,44 @@
...
+        )
+
+        for skill_data in race_data.get("skills", []):
+            skill, _ = Skill.objects.get_or_create(
+                name=skill_data["name"],
+                defaults={"bonus": skill_data["bonus"], "race": race}
+            )
+
+        guild = None
+        if player_data["guild"]:
```
Oleksl888, line 26: you can use walrus operator here `:=`

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done! Make sure you understand difference between different ondelete strategies

db/models.py
```diff
@@ -1 +1,38 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.PROTECT
```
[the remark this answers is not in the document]
mnezbrytska: No, I can't do it. I tried to do it before first pull recques and I got mistake.  I will write to you on Mate platform.

PR #958, opened 2024-08-27
Solution

AndriiHamasa:


db/models.py
```diff
@@ -1 +1,33 @@
 from django.db import models
+
+
+class Race(models.Model):
+    RACE_TYPES = (
+        ("Elf", "They have big and pointed ears"),
+        ("Dwarf", "They are short in stature and constantly mine gold"),
+        ("Human", "think they are smarter than others"),
+        ("Ork", "creepy freaks who want to kill everyone")
+    )
```
MNDonut, line 10: What if you get another file with other race types? It means migrations will fail because name field doesn't support other race types than you have here. To know if you can process a new file, you'll need to check it(it can be a large file with 100k rows) and add types here that is a bit odd and not unified

db/models.py
```diff
@@ -1 +1,33 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=True)
```
MNDonut, line 29: EmailField is unique by default

MNDonut (changes requested):

MNDonut (approved):

PR #960, opened 2024-08-27
solution

Sergi0bbb:


db/models.py
```diff
@@ -1 +1,35 @@
...
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=True)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="members"
```
oliasymak24, line 33: better say "players"
And be consistent, if you add related_name in one place, add also in others

main.py
```diff
@@ -1,10 +1,69 @@
 import init_django_orm  # noqa: F401
 
+import json
```
oliasymak24, line 3: Change the order of your imports according to `pep8`. Here `json` should be first.

main.py
```diff
@@ -1,10 +1,69 @@
...
+from django.utils import timezone
 from db.models import Race, Skill, Player, Guild
 
 
+def get_or_create_race(race_data: dict) -> Race:
+    race, created = Race.objects.get_or_create(
```
oliasymak24, line 9: If you don't use variables better put `race, _`
Check everywhere

main.py
```diff
@@ -1,10 +1,69 @@
...
+        guild: Guild
+) -> None:
+    Player.objects.get_or_create(
+        nickname=player_name,
+        defaults={
+            "email": player_data["email"],
+            "bio": player_data["bio"],
+            "race": race,
+            "guild": guild,
+            "created_at": timezone.now()
```
oliasymak24, line 38: You don't need to add time, as you added such parameter to model `auto_now_add`

oliasymak24 (changes requested):

MNDonut (approved):

PR #961, opened 2024-08-27
Solution

dpiuro:


main.py
```diff
@@ -1,10 +1,48 @@
 import init_django_orm  # noqa: F401
 
-from db.models import Race, Skill, Player, Guild
+from django.db import IntegrityError
+import json
```
oliasymak24, line 4: Change the order of your imports according to `pep8`. Here `json` should be first.

main.py
```diff
@@ -1,10 +1,48 @@
...
+from db.models import Race, Skill, Guild, Player
 
 
 def main() -> None:
-    pass
+    with open("db/tests/players.json") as file:
+        data = json.load(file)
+
+    for nickname, player_data in data.items():
+        race_data = player_data["race"]
```
oliasymak24, line 14: Use get here too)

db/models.py
```diff
@@ -1 +1,51 @@
...
+    def __str__(self) -> str:
+        return self.name
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
```
Oleksl888, line 17: Are you sure you have to delete a skill if some Race gets deleted? What if it's used by other Race?

> This is what checklist says:
> 2) Use everything for its intended purpose
> - Don't confuse the functionality of the on_delete parameter
> Use on_delete=models.CASCADE if you know that some model doesn't exist without other model. So when the referenced object is deleted, also delete the objects that have references to it.
> 
> Use on_delete=models.SET_NULL if you know that the object should not be deleted, even if the object it references may be deleted.

db/models.py
```diff
@@ -1 +1,51 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```
Oleksl888, line 6: As per checklist:

> description - a text field, can be blank.

would make sense to make this field null as well

main.py
```diff
@@ -1,10 +1,69 @@
...
+            race=race,
+            guild=guild
+        )
+    except IntegrityError:
+        print(f"Player with nickname {nickname} already exists.")
+
+
 def main() -> None:
-    pass
+    data = load_data("db/tests/players.json")
```
Oleksl888, line 64: Don't think you need to refer to test directory for this file. It should be in your main project dir

main.py
```diff
@@ -1,10 +1,69 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
+def load_data(filepath: str) -> dict:
+    with open(filepath) as file:
+        return json.load(file)
+
+
+def create_or_get_race(race_data: dict) -> Race:
+    race, _ = Race.objects.get_or_create(
```
Oleksl888, line 16: this function basically does one thing so it doesn't make much sense to create a funciton for it

Oleksl888: You can basically do this instead 
`race, _ = Race.objects.get_or_create(name=race_data["name"], description=race_data["description"])`

Oleksl888 (changes requested):

MNDonut (approved):

PR #962, opened 2024-08-27
Solution

Ivan-Shakhman:
i guess its it

### DIFF
--- a/db/models.py
+++ b/db/models.py
@@ -1,1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255, null=True, blank=True)
+    race = models.ForeignKey("Race", on_delete=models.CASCADE)
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey("Race", on_delete=models.CASCADE)
+    guild = models.ForeignKey("Guild", null=True , on_delete=models.SET_NULL)
+    created_at = models.DateTimeField(auto_now_add=True)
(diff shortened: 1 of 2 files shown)


db.sqlite3
Oleksl888, line 1: remove db please :dancers:

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255, null=True, blank=True)
+    race = models.ForeignKey("Race", on_delete=models.CASCADE)
```
Oleksl888, line 12: Why did you choose CASCADE here? It means that if Race gets deleted this skill will be gone too, but what if it's also referenced by other Race?

db/models.py
```diff
@@ -1 +1,26 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True, blank=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey("Race", on_delete=models.CASCADE)
```
Oleksl888, line 24: what about related names here?

main.py
```diff
@@ -1,10 +1,42 @@
 import init_django_orm  # noqa: F401
-
+import json
```
Oleksl888, line 2: there is certain imports order convention that you need to follow..

main.py
```diff
@@ -1,10 +1,42 @@
...
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as f:
+        players = json.load(f)
+
+        for name_player, data in players.items():
```
Oleksl888, line 10: so you execute all the below code while the file is open? That's not good

main.py
```diff
@@ -1,10 +1,42 @@
 import init_django_orm  # noqa: F401
-
+import json
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as f:
```
Oleksl888, line 7: `f` is not a descriptive name

Oleksl888 (changes requested):

MNDonut (approved):

PR #964, opened 2024-08-27
Solution

pashawarganov:


db/models.py
```diff
@@ -1 +1,38 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```

db/models.py
```diff
@@ -1 +1,38 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.SET_NULL, null=True)
```
Oleksl888, line 15: Good to see that someone made this right :+1:

db/models.py
```diff
@@ -1 +1,38 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(unique=True)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE, null=True)
```
Oleksl888, line 33: if this field can be NULL why would we use CASCADE then? Cascade means that this record cannot exist without this record that we are referring to. If you set this field as NULL it means that it CAN exist without race record reference. Think about it

db/models.py
```diff
@@ -1 +1,38 @@
...
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```
Oleksl888, line 34: This is correct - it does make sense

main.py
```diff
@@ -1,10 +1,44 @@
 import init_django_orm  # noqa: F401
+import json
```
Oleksl888, line 2: Please follow import guidelines - build-in libraries imports come first

main.py
```diff
@@ -1,10 +1,44 @@
...
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_file:
+        players = json.load(players_file)
+
+    for player, info in players.items():
+        race = info["race"]
+        guild = info["guild"]
+
+        new_race = Race.objects.get_or_create(
```
Oleksl888, line 15: Using indexes is not very `pythonic`. Since this method returns a tuple you can use unpacking like so: `new_race, _`
This will make your code more respectable by other developers :wink:

main.py
```diff
@@ -1,10 +1,44 @@
...
+            name=race["name"],
+            description=race["description"]
+        )[0]
+
+        if guild:
```
Oleksl888, line 20: Your code will be better structured if you set `guild` as `None` first and then use `if` statement once without an extra `else` instruction. It is a generally accepted approach not to use too much branching if possible. In some programming languages to many if's can even make your code slower, so try to use less if possible.
```
guild = None
if guild := player_data.get("guild"):
     ...

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: db/models.py (hunk 3 above)
Comment (Oleksl888) [inline db/models.py:33]: what about related names here?

File: db/models.py (hunk 1 above)
Comment (pashawarganov) [reply db/models.py:6]: I changed it, but in README we have: 'description - a text field, can be blank'. That`s a reason why I wrote only 'blank'

File: db/models.py [hunk 8]
```diff
@@ -1 +1,40 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.name}"
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(unique=True)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE,
```
Oleksl888, line 33: please make it pretty - each parameter on new line

main.py
```diff
@@ -1,11 +1,48 @@
...
+            email=info["email"],
+            bio=info["bio"],
+            race=new_race,
+            guild=new_guild
+        )
 
 
 if __name__ == "__main__":
     main()
+    # Race.objects.all().delete()
```
Oleksl888, line 45: remove comments please

db/models.py
```diff
@@ -1 +1,40 @@
...
+                             related_name="player_race")
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True,
+                              related_name="player_guild")
```
Oleksl888, line 36: This related name should be a link to `players` linked to guild or race. So what would be the most appropriate name here based on what you saw in examples?

Oleksl888 (changes requested):

db/models.py
```diff
@@ -1 +1,47 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(unique=True)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="race_players"
```
Oleksl888, line 36: ok:) but just `players` would be nice and simple

Oleksl888 (approved):
Well done! :+1:

db/models.py
pashawarganov: got it)

PR #966, opened 2024-08-27
Solution

Bogdan-Rylov:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Oleksl888, line 12: Why did you use CASCADE here? Does this mean Skill cannot exist without a Race?

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
```
Oleksl888, line 17: What about blank for this field?

db/models.py
```diff
@@ -1 +1,26 @@
...
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Oleksl888, line 24: related names here?

main.py
```diff
@@ -1,10 +1,37 @@
...
+import init_django_orm  # noqa: F401
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        players_dicts: dict[dict] = json.load(file)
+    for player_name, player_dict in players_dicts.items():
+        race = Race.objects.get_or_create(
```
Oleksl888, line 11: Tuple unpacking is preferable to using indexes: `race, _`

Oleksl888 (changes requested):

db/models.py
Bogdan-Rylov: Yes. As it is stated in the task, - "Each race has unique skills. Create a model Skill for them." -> no Race - no skills unique for this Race.

Oleksl888: Imagine you have skill A and it is used both by Race B and C. When race B gets deleted skill A is being deleted as well but race C still has reference to it.

Bogdan-Rylov: Yes, I understand this. It makes sense. I also thought about it. But in the task it is stated very clear: "Each race has **unique** skills". Regarding your assumption "Imagine you have skill A and it is used both by Race B and C": walking, fighting with a sword, shooting from a bow - are not unique skills, but rather abilities, judging from the logic of the task.
And, for example, an "elven special piercing bow shot" - is a skill unique to the Elf race, and it wouldn't make any sense, if an Ork race player could perform "elven piercing bow shot". Theoretically, it's possible that an ork player made friends with an elf NPC, and he taught him this skill. But it also depends on the genre of the game, - in MMO - it could be possible, in Action, Fighting, RTS - impossible.
Also imagine that developers decided that they don't want to have Elf race in their game anymore, because they changed lore. In this case, the ork player would still have "elven piercing bow shot". It contradicts the logic of the task.

oliasymak24 (approved):

PR #969, opened 2024-08-28
implement py-game-models

Maksym-Turenko:


db/models.py
```diff
@@ -1 +1,35 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```

db/models.py
```diff
@@ -1 +1,35 @@
...
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
```
Oleksl888, line 12: Are you sure you have to delete a skill if some Race gets deleted? What if it's used by other Race?

db/models.py
```diff
@@ -1 +1,35 @@
...
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="players"
+    )
+    guild = models.ForeignKey(
+        Guild,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="members"
```
Oleksl888, line 33: members is a good name actually but `players` would be more accurate

main.py
```diff
@@ -1,10 +1,49 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as f:
```
Oleksl888, line 9: `f` is not a very descriptive name

main.py
```diff
@@ -1,10 +1,49 @@
...
+        players_data = json.load(f)
+
+    for nickname, player_data in players_data.items():
+        race_name = player_data["race"]["name"]
```
Oleksl888, line 13: you don't really need to create extra variables for it since you use it only once

main.py
```diff
@@ -1,10 +1,49 @@
...
+        race_description = player_data["race"].get("description", "")
+        race, _ = Race.objects.get_or_create(
+            name=race_name,
+            defaults={"description": race_description}
+        )
+
+        for skill_data in player_data["race"].get("skills", []):
+            skill_name = skill_data["name"]
```
Oleksl888, line 21: same here - don't create extra variables - check in other places too

main.py
```diff
@@ -1,10 +1,49 @@
...
+            skill_bonus = skill_data["bonus"]
+            Skill.objects.get_or_create(
+                name=skill_name,
+                race=race,
+                defaults={"bonus": skill_bonus}
+            )
+        guild_data = player_data.get("guild")
```
Oleksl888, line 28: you can refactor this block without an `else`. First you can set guild as None and then use only one `if` to reset Guild if there is some data for it in json. Plus you can use walrus operator in your `if` statement.

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #971, opened 2024-08-31
Solution

Anatolik-ua:


main.py
```diff
@@ -1,10 +1,47 @@
...
+
+        Player.objects.get_or_create(
+            nickname=player_name,
+            defaults={
+                "email": player_data["email"],
+                "bio": player_data["bio"],
+                "race": race,
+                "guild": guild
+            }
+        )
```
vsmutok, line 44: it's better to just use `create` here if you don't need to get the result

main.py
```diff
@@ -1,10 +1,47 @@
...
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_file:
+        data = json.load(players_file)
+
+    print(data)
```
vsmutok, line 12: remove this `print`

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #979, opened 2024-09-16
Solution

ip-ilya:


db/models.py
```diff
@@ -1 +1,37 @@
...
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        related_name="players",
+        on_delete=models.CASCADE
+    )
+    guild = models.ForeignKey(
+        Guild,
+        related_name="players",
+        on_delete=models.SET_NULL, null=True
```
Dimosphen1, line 35: Place all parameters on a new line

db/models.py
```diff
@@ -1 +1,37 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE
```
Dimosphen1, line 14: Add `related_name` here as well

main.py
```diff
@@ -1,10 +1,41 @@
...
-    pass
+    with open("players.json") as json_config:
+        config = json.load(json_config)
+
+    for username, user_data in config.items():
+
+        race = Race.objects.get_or_create(
+            name=user_data["race"]["name"],
+            description=user_data["race"]["description"]
+        )[0]
```
Dimosphen1, line 16: Use unpacking instead of indexing here and in the other places as well
```suggestion
        race, _ = Race.objects.get_or_create(
            name=user_data["race"]["name"],
            description=user_data["race"]["description"]
        )
```

Dimosphen1 (changes requested):

db/models.py
ip-ilya: Initially, I added it, but got this error after running pytest -> https://postimg.cc/YhvSHMmQ
Pytest works fine without it

main.py
ip-ilya: It only worked with 'race '. 
With 'guild', I get the 'TypeError: cannot unpack non-iterable NoneType object' error

PR #982, opened 2024-09-18
Develop

FalseCoreCentury:


main.py
```diff
@@ -1,10 +1,43 @@
...
-
 from db.models import Race, Skill, Player, Guild
+import json
 
 
 def main() -> None:
-    pass
+    with (open("players.json", "r") as file):
+        players_data = json.load(file)
+        print(players_data)
```
Dimosphen1, line 9: Do not overload context manager as mentioned in the checklist

Dimosphen1, line 9: There is no need for a print, remove it

main.py
```diff
@@ -1,10 +1,43 @@
...
+                    defaults={"bonus": race_data.get("bonus", "")}
+                )
+
+            Player.objects.get_or_create(
+                nickname=nickname,
+                email=player_data["email"],
+                bio=player_data.get("bonus", ""),
+                race=race,
+                guild=guild,
+
```
Dimosphen1, line 39: There is no need for a blank line here, remove it

db/models.py
```diff
@@ -1 +1,53 @@
...
+    def __str__(self) -> CharField:
+        return self.name
+
+
+class Skills(models.Model):
+    name = models.CharField(max_length=255)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        related_name="skill_set",
```
Dimosphen1, line 18: Use another `related_name`, it should not be as default one

db/models.py
```diff
@@ -1 +1,53 @@
...
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+
+    def __str__(self) -> CharField:
+        return self.name
+
+
+class PlayerModel(models.Model):
```
Dimosphen1, line 34: It seems that the model name is incorrect

db/models.py
```diff
@@ -1 +1,53 @@
...
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        related_name="players_race",
```
Dimosphen1, line 40: Use changed model name in plural form as `related_name` for both FK fields

db/models.py
```diff
@@ -1 +1,53 @@
...
+        on_delete=models.CASCADE)
```
Dimosphen1, line 41: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue in all places

db/models.py
Dimosphen1, line 1: It is a good idea to add corresponding migration

Dimosphen1 (changes requested):
All tests should pass. Several changes were requested

PR #986, opened 2024-09-20
Develop

okolodka:


db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2024-09-20 21:07
```

main.py
```diff
@@ -1,10 +1,43 @@
...
+
+    for nickname, params in players.items():
+        params_race = params["race"]
+        race, _ = Race.objects.get_or_create(
+            name=params_race["name"],
+            description=params_race["description"],
+        )
+
+        params_skills = params_race["skills"]
+        for skill in params_skills:
```
pavlopro, line 19: Directly iterate over `params_race["skills"]`. No need to have this `params_skills` variable

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #991, opened 2024-09-25
solution

popeye88:


main.py
```diff
@@ -1,10 +1,43 @@
...
+    data = json.load(open("players.json", "r")).items()
+
+    for player_name, player_data in data:
+        race_data = player_data["race"]
+        guild_data = player_data["guild"]
+
+        race = Race.objects.get_or_create(
+            name=race_data["name"],
+            defaults={"description": race_data["description"]}
+        )[0]
```
Dimosphen1, line 17: Please, use unpacking instead of retrieving value by index here and in the other places as well

db/migrations/0002_alter_skill_name.py
Dimosphen1, line 1: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

main.py
```diff
@@ -1,10 +1,43 @@
 import init_django_orm  # noqa: F401
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    data = json.load(open("players.json", "r")).items()
```
Dimosphen1, line 8: Please, use context manager here to work with file with an implementation tips from the checklist

Dimosphen1 (changes requested):

main.py
```diff
@@ -1,10 +1,44 @@
...
+        )
+
+        for skill_data in player_data["race"]["skills"]:
+            Skill.objects.get_or_create(
+                name=skill_data["name"],
+                defaults={"bonus": skill_data["bonus"]},
+                race=race,
+            )
+
+        if guild_data:
```
pavlopro, line 27: Define `guild = None` first

main.py
```diff
@@ -1,10 +1,44 @@
...
 def main() -> None:
-    pass
+    with open("players.json") as file:
+        data = json.load(file).items()
+
+    for player_name, player_data in data:
+        race_data = player_data["race"]
+        guild_data = player_data["guild"]
+
+        race, created = Race.objects.get_or_create(
```
pavlopro, line 15: no need to have this `created` variable, use `_` instead

main.py
```diff
@@ -1,10 +1,44 @@
...
+
+        for skill_data in player_data["race"]["skills"]:
+            Skill.objects.get_or_create(
+                name=skill_data["name"],
+                defaults={"bonus": skill_data["bonus"]},
+                race=race,
+            )
+
+        if guild_data:
+            guild, created = Guild.objects.get_or_create(
```
pavlopro, line 28: here the same

pavlopro (changes requested):

main.py
popeye88: Pls, give me some explanation. Do not understand

PR #994, opened 2024-10-01
The solution of the task

Ja0ni:


db/models.py
```diff
@@ -1 +1,42 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
pavlopro, line 9: No need f string

pavlopro (approved):
Good Job!

PR #1000, opened 2024-10-12
Solution

PavelTym:


main.py
```diff
@@ -1,10 +1,46 @@
...
+            )
+
+        Player.objects.get_or_create(
+            nickname=player_name,
+            defaults={
+                "email": player_info["email"],
+                "bio": player_info["bio"],
+                "race": race,
+                "guild": guild,
+                "created_at": now(),
```
pavlopro, line 41: No need to have `"created_at": now()` here

pavlopro (changes requested):
Please check tests
They failed

main.py
PavelTym: тести в пайчарм всі проходять а тут чомусь ні

PavelTym:
тести в пайчарм всі проходять а тут чомусь ні

pavlopro (changes requested):
Tests failed

pavlopro (changes requested):
It is a good practice to include one migration per app per PR in Django projects since it helps to avoid migrations folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

PR #1010, opened 2024-10-28
Develop

KKvladys:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
```
vladyslav-tmf, line 5: As task requires:
Each player should choose a race to play, such as Elf, Dwarf, Human, or Ork.

So, you need to add argument `choices` to `name` field.

db/models.py
```diff
@@ -1 +1,26 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=False)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now=True)
```
vladyslav-tmf, line 26: From checklist:
- Don't confuse `auto_now` and `auto_now_add`

When you use `auto_now_add` - it records the time and date of creation.
When you use `auto_now` - updates the value of the field to the current time and date every time when the object changes.

Here you need to use `auto_now_add` parameter.

vladyslav-tmf (changes requested):
Your code is ok, you just need to fix a few minor issues.

db/models.py
```diff
@@ -1,8 +1,14 @@
 from django.db import models
 
+RACES_NAME = [
+    ("Elf", "Elf"),
+    ("Human", "Human"),
+    ("Dwarf", "Dwarf")
```
vladyslav-tmf, line 6: You forgot to add Ork race.

vladyslav-tmf (approved):
Good!

PR #1011, opened 2024-10-28
Solution

VladyslavBon:


db/models.py
```diff
@@ -1 +1,40 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```
KKvladys, line 6: As task requires:
Each player should choose a race to play, such as Elf, Dwarf, Human, or Ork.

So, you need to add argument choices to name field.

KKvladys (approved):

db/models.py
Nattalli: also having RACES_NAME inside of the Race class would be better

db/models.py
```diff
@@ -1 +1,48 @@
...
+    name = models.CharField(max_length=255, unique=True, choices=RACES_NAME)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        related_name="skill_set",
```
Nattalli, line 22:
```suggestion
        related_name="skills",
```

Nattalli (changes requested):

db/models.py
VladyslavBon: Tests break if it will be 'skills', I was try it first

AnyoneClown (changes requested):
Make sure, that all tests passes before asking for a review. Check comments above that you didn't fix

main.py
```diff
@@ -1,10 +1,43 @@
...
+        if guild:
+            guild = Guild.objects.get_or_create(**guild)[0]
+
+        Player.objects.create(
+            nickname=person,
+            email=player.get("email"),
+            bio=player.get("bio"),
+            race=race,
+            guild=guild
+        )
```
Arsen-hrynevych, line 40: It would be better to split it up into smaller functions

Arsen-hrynevych (approved):
ok

PR #1014, opened 2024-10-31
game-models solution

94nj111:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255)
```
vladyslav-tmf, line 5: As task requires:
Each player should choose a race to play, such as `Elf`, `Dwarf`, `Human`, or `Ork`.
You should use the `choices` option to define the possible races for players to select from.

vladyslav-tmf (changes requested):

PR #1017, opened 2024-11-01
Solution

Atikiho:


main.py
```diff
@@ -1,11 +1,37 @@
...
+            name=player_data["race"]["name"],
+            defaults={"description": player_data["race"]["description"]}
+        )
+        skills = []
+        for skill_data in player_data["race"]["skills"]:
+            skill, _ = Skill.objects.get_or_create(
+                name=skill_data["name"],
+                defaults={"bonus": skill_data["bonus"], "race": race}
+            )
+            skills.append(skill)
```
Mxbely, line 22: You do not need save skiils

main.py
```diff
@@ -1,11 +1,37 @@
...
+        else:
+            guild = None
+        player = Player(
+            nickname=player_name,
+            email=player_data["email"],
+            bio=player_data["bio"],
+            race=race,
+            guild=guild,
+        )
+        player.save()
```
Mxbely, line 37: you can use get_or_create for create player and don't use variables

Mxbely (changes requested):
Separating the logic creation function would be improved readability, and get_or_create can raise exceptions, so would be better use try/except

Atikiho:
> Separating the logic creation function would be improved readability, and get_or_create can raise exceptions, so would be better use try/except

What kind of exceptions? I thinked that all exceptions that should be handled are already handled inside this function

PR #1020, opened 2024-11-03
Solution-py-game-models

skyfoxwork:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
```
vladyslav-tmf, line 5: > Each player should choose a race to play, such as Elf, Dwarf, Human, or Ork.

Your `Race` model currently doesn't include predefined choices like `Elf`, `Dwarf`, `Human`, or `Ork`, as mentioned in the task. Consider adding a `choices` parameter to ensure players can select from these specific options.

main.py
```diff
@@ -1,10 +1,67 @@
...
+            guild_description = guild_info.get("description")
+
+            guild, guild_created = Guild.objects.get_or_create(
+                name=guild_name,
+                description=guild_description
+            )
+        else:
+            guild = None
+
+        Player.objects.create(
```
vladyslav-tmf, line 50: You should use `Player.objects.get_or_create`(...) instead of `Player.objects.create`(...) to avoid error if you try to create a player with an existing nickname, since the nickname field is unique (`unique=True`).

vladyslav-tmf (changes requested):
Your code is good, you just need to fix a few minor issues.

PR #1025, opened 2024-11-04
Solution

MrEgorLite:


db/models.py
```diff
@@ -1 +1,26 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
```
VladyslavBon, line 6: Task requires that each player should choose a race to play, such as Elf, Dwarf, Human, or Ork.
You have to create variable RACES_NAME inside of the Race class and add argument choices to name field.

db/models.py
```diff
@@ -1 +1,26 @@
...
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(unique=False)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
```
VladyslavBon, line 25: In checklist they ask us add related_name argument.

PR #1038, opened 2024-11-06
'Solution'

yulia-pl:


db/models.py
```diff
@@ -1 +1,30 @@
...
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race,
+        on_delete=models.CASCADE,
+        related_name="skill_set"
```
Arsen-hrynevych, line 15: By default, Django will create the name skill_set if you don't add a related_name. If you add a related_name, it should be something like skill rather than skill_set.

main.py
```diff
@@ -1,10 +1,45 @@
...
+
+        Player.objects.get_or_create(
+            nickname=player_name,
+            defaults={
+                "email": player_data["email"],
+                "bio": player_data["bio"],
+                "race": race,
+                "guild": guild
+            }
+        )
```
Arsen-hrynevych, line 42: Separate into a few small functions.

Arsen-hrynevych (changes requested):
Don't forget to check the task checklist.

> Do not forget to add related_name to ForeignKey fields

db/models.py
yulia-pl: Ok, I understand the issue with related_name. However, according to the tests, I can't set something like "skill" because the default value is specified in the tests. Therefore, in this case, I don't see anything better than simply removing the line with related_name="skill_set" so that Django creates it with the default value.
Here is a snippet of the test with skill_set: 
assert list(Race.objects.get(name="elf").skill_set.values_list("name"))

Nattalli: that is because tests use default calling

Nattalli (approved):

PR #1041, opened 2024-11-06
solution

dxrrkwm:


db/models.py
```diff
@@ -1 +1,30 @@
...
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True)
+
+
+class Skill(models.Model):
+    name = models.CharField(max_length=255)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(
+        Race, related_name="skill_set", on_delete=models.CASCADE
```
Nattalli, line 13: `related_name="skills"` would be better

Nattalli (approved):

PR #1044, opened 2024-11-08
Student

380974261314kv:


db/models.py
```diff
@@ -1 +1,47 @@
...
+
+class Skill(models.Model):
+    """Model describes skills for each race"""
+
+    name = models.CharField(max_length=255, unique=True)
+    bonus = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+
+    def __str__(self) -> str:
+        return str(self.name)
```
pavlopro, line 22: No need `str` here

db/models.py
```diff
@@ -1 +1,47 @@
...
+
+
+class Guild(models.Model):
+    """Guild model"""
+
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+    def __str__(self) -> str:
+        return str(self.name)
```
pavlopro, line 32: here as well

main.py
```diff
@@ -1,10 +1,36 @@
 import init_django_orm  # noqa: F401
```
pavlopro, line 1: add blank line between different import groups

pavlopro (changes requested):

PR #1060, opened 2024-11-21
implement py-game-models

yuliiapyrih:


main.py
```diff
@@ -1,10 +1,49 @@
...
+        player_skills = player_race["skills"]
+        player_guild = player_data["guild"]
+
+        race, _ = Race.objects.get_or_create(
+            name=player_race["name"],
+            defaults={"description": player_race["description"]}
+        )
+
+        for skill_data in player_skills:
+            skill, _ = Skill.objects.get_or_create(
```
Nattalli, line 23: btw you don't need to save object to the variable
```suggestion
            Skill.objects.get_or_create(
```

Nattalli (approved):

PR #1109, opened 2025-01-08
Solution

IvankaKuzin:


main.py
```diff
@@ -1,10 +1,47 @@
...
+                email=player_info["email"],
+                bio=player_info["bio"],
+                race=some_race
+            )
+
+        else:
+            some_guild, created = Guild.objects.get_or_create(
+                name=player_info["guild"]["name"],
+                description=player_info["guild"]["description"],
+            )
```
taraskhom, line 36: You can check if a guild exists, create it, or pass a null value. This will allow you not to repeat the code for creating a player, because you will take it out of a guild verification

taraskhom (changes requested):
Well done but you can improve your solution

PR #1117, opened 2025-01-09
Solution

Arsenmyron:


bodiakof (approved):
Not too shabby!

Although, you could've applied suggestions from the [Checklist](https://github.com/mate-academy/py-game-models/blob/master/checklist.md) like these:

```
Good example (.get() method returns None by default):

guild = data.get("guild")
if guild:
    ...
Bad example:

if "guild" in data:
    guild = data["guild"]
```

PR #1119, opened 2025-01-09
Solution_game-models

SummerKatrin:


ihorhalyskiy (approved):
i approve but i think we need only 2 migration, ask mentors pls about this

PR #1143, opened 2025-01-23
project_

apmaz:


main.py
```diff
@@ -1,10 +1,47 @@
...
+    Race.objects.all().delete()
+    Skill.objects.all().delete()
+    Guild.objects.all().delete()
+    Player.objects.all().delete()
+
+    with open("players.json", "r") as f:
+        data = json.load(f)
+
+    players = data
+    for player in players:
```
vsmutok, line 18: Here you can immediately iterate over both the key and the value, which will simplify your code a bit.

main.py
```diff
@@ -1,10 +1,47 @@
...
+
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    Race.objects.all().delete()
+    Skill.objects.all().delete()
+    Guild.objects.all().delete()
+    Player.objects.all().delete()
```
vsmutok, line 12: use loop here, for example :

```python
for model in (Race, Skill, ...):
    model.objects.all().delete()
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1154, opened 2025-01-31
1

MikolaRozhansky:


db/models.py
```diff
@@ -1 +1,27 @@
...
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
```
SerhiyTo, line 24: Use `TextField` instead of `ChatField` in your case

SerhiyTo (changes requested):

db/models.py
MikolaRozhansky: У завданні ж написано що треба char field.

MikolaRozhansky: При чьому тут TextField ?

SerhiyTo: Ok, but, in the future, if this field must contain a maximum number of characters of 255, then it is better to consider using a TextField instead of a CharField

SerhiyTo (approved):

PR #1163, opened 2025-02-07
Created all models

shafran-andriy:


db/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.0.2 on 2025-02-05 23:48
```
SerhiyTo: Make sure you have only 1 migration file per pull request. It's not a good practice to make many migration files in one iteration

SerhiyTo (changes requested):

db/migrations/0001_initial.py
shafran-andriy: Of course, thanks for the advice, but during the study there was nowhere indicated in the materials that only one file should be in the migrations

shafran-andriy: fixed

main.py
```diff
@@ -1,10 +1,34 @@
...
+
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as players_file:
+        players = json.load(players_file)
+    for player in players:
```
vsmutok:
```suggestion
    for player_name, player_data in players.items():
```
This will allow you to simplify the code a bit))

vsmutok (changes requested):

main.py
shafran-andriy: I don`t need items because i use only keys in the dist. And as we know by default in dict user function keys()

vsmutok: This will help simplify your code. Instead of always using `players[player]`, you can use `player_data`, which is a better and more efficient approach.

vsmutok (changes requested):

shafran-andriy:
@vsmutok Thank you very much for the explanation

vsmutok (approved):
Great!

PR #1177, opened 2025-02-18
Solution

Baranotik15:


db/models.py
```diff
@@ -1 +1,28 @@
 from django.db import models
+
+
+class Race(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(blank=True, null=True)
```
Y-Havryliv: it's better not to use `blank=True` and `null=True` for string-based fields at the same time, `blank=True` will be enough. Fix everywhere

db/models.py
```diff
@@ -1 +1,28 @@
...
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(
+        Guild, on_delete=models.SET_NULL, null=True, blank=True
```
Y-Havryliv: `null` should be enough here. Remind yourself the difference between `null` and `blank`

main.py
```diff
@@ -1,10 +1,47 @@
...
+        if player_data.get("guild"):
+            guild, _ = Guild.objects.get_or_create(
+                name=player_data["guild"]["name"],
+                defaults={
+                    "description": player_data["guild"].get("description", "")
+                }
+            )
+
+        Player.objects.get_or_create(
+            nickname=player_name,  # Используем ключ JSON как nickname
```
Y-Havryliv: delete redundant comments

Y-Havryliv (changes requested):

Baranotik15:
sorry i misklik

main.py
oliasymak24, line 1: For better readability will be good to split into functions

main.py
```diff
@@ -1,10 +1,47 @@
 import init_django_orm  # noqa: F401
+import json
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as f:
```
oliasymak24: `f` is not a good name for variable

oliasymak24 (changes requested):

Y-Havryliv (approved):

PR #1183, opened 2025-02-23
Solution

kram3ko:


db/models.py
```diff
@@ -1 +1,27 @@
...
+
+
+class Guild(models.Model):
+    name = models.CharField(max_length=255, unique=True)
+    description = models.TextField(null=True)
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255, unique=False)
```
oliasymak24: `unique` by default is False, so no sense to set it

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1185, opened 2025-02-23
'Solution'

JohnyJDX:


main.py
```diff
@@ -1,10 +1,44 @@
...
+        player_guild = data_player["guild"]
+        player_skills = player_race["skills"]
+
+        race, _ = Race.objects.get_or_create(
+            name=player_race["name"],
+            defaults={"description": player_race["description"]}
+        )
+
+        guild = None
+        if player_guild:
```
Nattalli:
```suggestion
        if player_guild := data_player["guild"]:
```

Nattalli: the same everywhere so you don't use a memory for data that would be used for once

Nattalli (changes requested):

PR #1205, opened 2025-03-11
Solution

Revko:


db/migrations/0002_alter_player_guild.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 5.1.7 on 2025-03-11 16:01
```
SerhiyTo: I strongly recommend creating only 1 migration per app within a single PR. This is a very good practice

SerhiyTo (changes requested):

SerhiyTo (changes requested):
Did you forget to delete the file with the second migration?

main.py
```diff
@@ -1,10 +1,48 @@
...
+
+        Player.objects.get_or_create(
+            nickname=nickname,
+            defaults={
+                "email": player["email"],
+                "bio": player["bio"],
+                "race": race,
+                "guild": guild,
+            }
+        )
```
vsmutok: If you don't need to get the result, use **create** instead of **get_or_create**.

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1234, opened 2025-03-23
solution

DHlomozda:


Claymore-AI (approved):
good job

DHlomozda:
@mateacademy-ai-mentor, I have two question for your last comment.
-I already have in Player model two fields which is foreing key to another model. These two fields have parametr on_delete. For rese field on_delete=models.CASCADE and models.SET_NULL for guild field. What do you mean " Specifically, the on_delete parameter is missing in the Player model"?
-what do you want me to change in related_name? Do you not like names in this field?

PR #1363, opened 2025-07-11
Completed game models and data import with tests passed

kravetst:


kravetb (approved):
Please, don't use chat gbt for complete all task!

PR #1384, opened 2025-07-28
Implemented various models in order to add records to DB from json file.

Otst0y:


main.py
```diff
@@ -1,10 +1,47 @@
+import json
+
 import init_django_orm  # noqa: F401
 
 from db.models import Race, Skill, Player, Guild
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r+") as file:
```
dimashevchukk: only reading is used, r+ not needed

main.py
```diff
@@ -1,10 +1,47 @@
...
+        data = json.load(file)
+
+    for player_name, player_info in data.items():
+        race_description = player_info["race"]["description"] \
+            if player_info["race"]["description"] else ""
```
dimashevchukk: use get with default ""

main.py
```diff
@@ -1,10 +1,47 @@
...
+
+        for skill in skills:
+            Skill.objects.get_or_create(
+                name=skill.get("name"),
+                bonus=skill.get("bonus"),
+                race=race_obj
+            )
+
+        guild = player_info.get("guild")
+        guild_obj, created = (
```
dimashevchukk: created is not used so better change it to placeholder _

dimashevchukk (changes requested):

dimashevchukk (approved):

PR #1387, opened 2025-07-28
Solution

DenisPrudius:


main.py
```diff
@@ -1,10 +1,48 @@
...
+            )
+
+        Player.objects.get_or_create(
+            nickname=nickname,
+            defaults={
+                "email": player["email"],
+                "bio": player["bio"],
+                "race": race,
+                "guild": guild,
+                "created_at": now()
```
yagginov, line 43: You don't need to write `"created_at": now()` because you've already defined it at the model level.

yagginov (approved):
well done

PR #1394, opened 2025-07-29
created models, implemented main func

Sashoxxx:


VadymKyrylko (approved):
Good work! Maybe better to use get_or_create also in Player instead of create?

PR #1502, opened 2025-11-23
Django Game-Models

n-korobko:


db/models.py
```diff
@@ -1 +1,35 @@
...
+        return self.name
+
+
+class Player(models.Model):
+    nickname = models.CharField(max_length=255, unique=True)
+    email = models.EmailField(max_length=255)
+    bio = models.CharField(max_length=255)
+    race = models.ForeignKey(Race, on_delete=models.CASCADE)
+    guild = models.ForeignKey(Guild, on_delete=models.SET_NULL, null=True)
+    created_at = models.DateTimeField(auto_now_add=True)
```
stanislav-kishechko, line 48: You forget to add `related_name` to ForeignKey fields. It will allow you to have access to related models on Many side

stanislav-kishechko (approved):

db/models.py
n-korobko: Thanks, i do it

main.py
```diff
@@ -1,10 +1,50 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r") as file:
+        players_data = json.load(file)
+
+    for nickname, data in players_data.items():
+        email = data["email"]
+        bio = data["bio"]
```
petrykivd, line 13: What will happen if key "email" does not exists in dict data? Better use .get for retrieving values from dict by key.

petrykivd: Fix it in a whole code

petrykivd (changes requested):

PR #1674, opened 2026-04-16
solved

pawunder1992:


main.py
```diff
@@ -1,10 +1,38 @@
...
+        Race.objects.get_or_create(name=race_name,
+                                   description=race_description)
+        guild_name = data[player]["guild"]
+        if guild_name:
+            Guild.objects.get_or_create(name=(guild_name.get("name")),
+                                        description=guild_name
+                                        .get("description")
+                                        if guild_name else None)
+
+        race = Race.objects.get(name=data[player]["race"]["name"])
```
VadimPusiak788: You're calling Race.objects.get_or_create() and then immediately calling Race.objects.get() again, which causes an unnecessary extra database query. You can unpack the object returned by get_or_create() instead.
`race, created = Race.objects.get_or_create(name=race_name, description=race_description)`

main.py
```diff
@@ -1,10 +1,38 @@
...
+        if guild_name:
+            guild = Guild.objects.get(name=guild_name.get("name"))
```
VadimPusiak788: Same issue as above — get_or_create already returns the object, no need to query again.

main.py
```diff
@@ -1,10 +1,38 @@
...
+    for player in data:
+        race_name = data[player]["race"]["name"]
+        race_description = data[player]["race"]["description"]
+        Race.objects.get_or_create(name=race_name,
+                                   description=race_description)
+        guild_name = data[player]["guild"]
+        if guild_name:
+            Guild.objects.get_or_create(name=(guild_name.get("name")),
+                                        description=guild_name
+                                        .get("description")
```
VadimPusiak788: get() returns None by default, so you don't need to add an else condition.

VadimPusiak788 (changes requested):

main.py
pawunder1992: done

VadimPusiak788 (approved):

PR #1691, opened 2026-04-28
Solution

vova-dev-python:


pawunder1992 (changes requested):
fix the problem, i think you need to make migrations

PR #1693, opened 2026-04-29
Solution

lubkoserhii:


alisahoys (approved):
Proper use of get_or_create, Good structure

PR #1699, opened 2026-04-30
Implement models and main function

Prus-Oleksandr:


alisahoys (approved):
Solid logic, correct ORM usage

PR #1714, opened 2026-05-07
Implemented main() func to transport data from json file to db

Neyrony:


Neyrony:
Why did you give feedback about the migrations files, it's unnecesary it was created by Django

PR #1800, opened 2026-08-13
Develop

ktm2142:


main.py
```diff
@@ -1,10 +1,40 @@
...
 
 
 def main() -> None:
-    pass
+    with open("players.json", "r", encoding="utf-8") as file:
+        players_data = json.load(file)
+
+    for data, values in players_data.items():
+        guild = None
+        if values["guild"] is not None:
```
Sn1kls: [Checklist #5](https://github.com/mate-academy/py-game-models/blob/master/checklist.md) – use `.get()` for dict access. `values["guild"]` works only because every player has the key; use `values.get("guild")` instead: `if values.get("guild") is not None:`

Sn1kls (changes requested):

Sn1kls (approved):
Approved. Nice work!

PR #1802, opened 2026-08-14
creating-game-models-with-json-format

Dr-Kate:


db/migrations/0002_alter_player_guild_alter_player_race_and_more.py
Sn1kls, line 1: Optional (non-blocking): this second migration only adds `related_name` to the FKs, which means they weren't in the initial migration. It works fine as-is, but for a cleaner history you could regenerate a single initial migration that already contains everything: delete `0001` and `0002`, then run `python manage.py makemigrations` so `0001_initial.py` includes the `related_name`s from the start. Purely a tidiness thing.

Sn1kls (approved):
Approved. One optional cleanup note inline – non-blocking)

